PR #44, opened 2014-07-28
Make jline logging configurable

myaut:
Current implementation of JLineShell forces all logs being written to console, regardless of externally provided configuration. Since it is done in a separate thread, there is no way to revert settings cleaner.

I've added an configuration option that disables such behavior (probably may be done using cleaner fix)


ericbottard:
Closing outdated PRs and issues

PR #61, opened 2015-02-09
SHL-170: Allow space-text for values

ericbottard:
Make mandatory handling code more readable. Also allow for the empty key to not be the first one


ericbottard:
Merged as 511a7e2 with added tests

PR #67, opened 2015-03-19
SHL-174: Multi-Step Commands

robinhowlett:
- Created a new @CliStepIndicator annotation to denote @CliCommand methods that have multiple steps
- Created a new abstract class AbstractStepExecutionProcessor for permitting step-level control of commands
- Added StepCommandsTest demonstrating the feature

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,7 +3,7 @@ buildscript {
         maven { url 'http://repo.springsource.org/plugins-release' }
     }
     dependencies {
-        classpath 'org.springframework.build.gradle:docbook-reference-plugin:0.1.5'
+        classpath 'io.spring.gradle:docbook-reference-plugin:0.3.0'
         classpath 'org.springframework.build.gradle:spring-io-plugin:0.0.3.RELEASE'
     }
 }
--- a/src/main/java/org/springframework/shell/core/AbstractShell.java
+++ b/src/main/java/org/springframework/shell/core/AbstractShell.java
@@ -300,4 +300,11 @@ public abstract class AbstractShell extends AbstractShellStatusPublisher impleme
 			logger.info(result.toString());
 		}
 	}
+
+	/**
+	 * @return the logger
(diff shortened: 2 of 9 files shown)


ericbottard:
Overall, I'm not too keen on adding this to the core of Spring Shell.

It is a rather specific solution to a specific problem and can be implemented without changing the core IMO (there is nothing preventing one from having command classes interact with some shared state that "knows" where we stand in a multi-step process). This could even leverage @CliAvailabilityIndicator.

Also, the provided example (the test) does nor really showcase how the added framework methods
```
stepConfig = configureStep(stepResult);
stepResult = executeStep(stepResult, stepConfig);
handleStepExecutionResult(invocationContext, stepResult);
```
are useful. Maybe a more real-world usecase would enlighten me.

robinhowlett:
@ericbottard I understand your concerns and yes the test is a bit simple to demonstrate the feature.

The main reason I requested/provided this feature was that I was having some difficultly implementing a "user menu"-style of interaction with commands, specifically around commands that were returning data that had natural "next" or "previous" steps that I wanted to allow the user to choose to navigate to.

To show a more real-world use case, I created [Mockbin CLI](https://github.com/robinhowlett/mockbin-cli), a Spring Shell application that interacts with [Mockbin](http://mockbin.com/). 

The below GIF demonstrates an example use of a multi-step command. The command executes an request to Mockbin which returns a list of objects, prints those objects to an ASCII table and prompts the user to either enter the row number of an entry to be displayed in JSON, re-print the table, or exit the command execution. (The multiple output format support is part of [SHL-175](https://github.com/spring-projects/spring-shell/pull/68)).

CLI Example

ericbottard:
Closing outdated PRs and issues

MingningShao:
A menu style interactive Command tool was quite useful.
Lots of user need this function.
Hope the spring-shell can support it.

PR #78, opened 2015-07-01
The static getter for those ThreadLocal bound objects are never used.

ericbottard:
Will close SHL-88 as well, obviously.


markpollack:
I think this was related to some functionality that users used in Roo when creating some commands.  Will investigate.

markpollack:
I thought it might be related to this 'focus' option - https://jira.spring.io/browse/ROO-784  but after grepping the roo code base for any other use of the thread local, I found no usage.  Looks ok to remote.

ericbottard:
Merged as 73f47a4

PR #82, opened 2015-07-18
cosmetic change

prashant1sharma:


ericbottard:
Contrary to some languages (like French), it is usual in English to have a space after a colon, but not before.

PR #86, opened 2015-10-19
Add pretty table rendering

ericbottard:
This is work I have been sitting on for a while.

It adds a rich Table rendering framework, supporting multiline rendering, text-aligning, nice border styles, text-wrapping, adaptive terminal-width support, etc.

The entry point is class [Table](https://github.com/ericbottard/spring-shell/blob/5efaa3d75349907bfba4dbc073b3d3487a93e21d/src/main/java/org/springframework/shell/table/Table.java), having a look at its javadoc is a good idea.

Here is a demo of the new Tables in action, using Spring XD as an example (there is some rendering artifact in the screen capture that does not appear in a real terminal):

[asciicast](https://asciinema.org/a/0b1lrrbr94tzlbkbawy1wchxt)


src/main/java/org/springframework/shell/core/AbstractShell.java
```diff
@@ -141,6 +144,7 @@ public CommandResult executeCommand(String line) {
 			return new CommandResult(true, result, null);
 		} catch (RuntimeException e) {
 			setShellStatus(Status.EXECUTION_FAILED, line, parseResult);
+			e.printStackTrace();
```
markpollack: use logger.warning?

ericbottard: good catch.

src/main/java/org/springframework/shell/table/Table.java
```diff
@@ -0,0 +1,389 @@
...
+
+		aligners.put(CellMatchers.table(), left);
+
+	}
+
+	public TableModel getModel() {
+		return model;
+	}
+
+	public Table withBorder(int top, int left, int bottom, int right, int match, BorderStyle style) {
```
markpollack: addBorderSpecification or addBorder would be a better method name.

src/main/java/org/springframework/shell/table/Table.java
```diff
@@ -0,0 +1,389 @@
...
+
+		// TABLE BOTTOM RIGHT CORNER
+		borders.paintCorner(rows, columns, result);
+		if (result.length() > before) {
+			result.append('\n');
+		}
+		return result.toString();
+	}
+
+	private int[] computeActualColumnWidths(int availableWidth, int[] minCellWidths, int[] maxCellWidths) {
```
markpollack: computeColumnWidths seems sufficient as a name.

src/main/java/org/springframework/shell/table/Table.java
```diff
@@ -0,0 +1,389 @@
...
+		} // Redistribute nicely
+		else {
+			int W = availableWidth - minTableWidth;
+			int D = maxTableWidth - minTableWidth;
+			cellWidths = new int[columns];
+			for (int column = 0; column < columns; column++) {
+				cellWidths[column] = minCellWidths[column] + W * (maxCellWidths[column] - minCellWidths[column]) / D;
+			}
+		}
+		// TODO: handle rounding error?
```
markpollack: Create a JIRA for it vs. TODO

src/main/java/org/springframework/shell/table/Table.java
```diff
@@ -0,0 +1,389 @@
...
+		return cellWidths;
+	}
+
+	public Table format(CellMatcher cells, Formatter formatter) {
```
markpollack: looking at these methods that help build the table via a fluent API makes me think that there should be a TableBuilder class.  Then 'withBorder' 'withAligment', 'withSize' would more natural method names in the context of a builder class.

markpollack: Also a builder class looks like it could hide the repetition of 

```
		table.format(CellMatchers.ofType(Map.class), new MapFormatter(" = "));
		table.align(CellMatchers.ofType(Map.class), new KeyValueHorizontalAligner("="));
		table.size(CellMatchers.ofType(Map.class), new KeyValueSizeConstraints(" = "));
		table.wrap(CellMatchers.ofType(Map.class), new KeyValueTextWrapper());```

passing in the same `CellMatchers` instance over and over again.

Comment (ericbottard) [inline src/main/java/org/springframework/shell/table/Table.java]: Well, 'Table' ils the builder really. A table is nothing until it is
render()ed.

File: src/main/java/org/springframework/shell/table/ArrayModel.java [hunk 6]
```diff
@@ -0,0 +1,49 @@
...
+package org.springframework.shell.table;
+
+import org.springframework.util.Assert;
+
+/**
+ * A TableModel backed by a row-first array.
+ *
+ * @author Eric Bottard
+ */
+public class ArrayModel extends TableModel {
```
markpollack: `DefaultTableModel` or `ArrayTableModel` ?

src/main/java/org/springframework/shell/table/BeanListModel.java
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * A table model that is backed by a list of beans.
+ *
+ * <p>One can control which properties are exposed (and their order). There is also
+ * a convenience constructor for adding a special header row.</p>
+ *
+ * @author Eric Bottard
+ */
+public class BeanListModel<T> extends TableModel {
```
markpollack: `BeanTableModel` ?

src/main/java/org/springframework/shell/table/KeyValueTextWrapper.java
```diff
@@ -0,0 +1,69 @@
...
+
+package org.springframework.shell.table;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * A TextWrapper implementation tailored for key-value rendering (working in concert
+ * with {@link KeyValueSizeConstraints}, {@link KeyValueHorizontalAligner}), that tries its
+ * best to vertically align some delimiter character (defautl '=').
```
markpollack: typo 'default`

src/main/java/org/springframework/shell/table/AssertingAlignmentStrategy.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.Arrays;
+
+import org.springframework.util.Assert;
+
+/**
+ * A decorator AlignmentStrategy, useful for debugging.
+ *
+ * @author Eric Bottard
+ */
+public class AssertingAlignmentStrategy implements AlignmentStrategy {
```
markpollack: `DebugAlignmentStrategy` ?

src/main/java/org/springframework/shell/table/AlignmentStrategy.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package org.springframework.shell.table;
+
+/**
+ * A strategy interface for performing text alignment.
+ *
+ * @author Eric Bottard
+ */
+public interface AlignmentStrategy {
```
markpollack: maybe `Aligner` instead vs. putting strategy in the interface name (seems not necessaary) as most subclasses naturally use the `Aligner` suffix.

src/main/java/org/springframework/shell/table/Table.java
markpollack: Table has 8 private variables and a public constructor - which is pretty much what a normal class would have.  So the class itself isn't just doing building, it is the state plus constructor plus configuraitons methods (which look oddly named outside a pure builder only class)

ericbottard:
@markpollack Addressed basic review comments. Still not sure about the Builder aspect of things.

The "state" of the Table is really 2 things: the model (which is a simple reference) and a bunch of render-influencing policies. There is no intermediate form between the model and the rendered String.

We could introduce _e.g._ a `TableCustomizer` that would allow chained customizations on a single `CellMatcher`, but there would be little gain IMO (that KeyValue setup is a bit extraordinary as many customizations need to apply at once, but this is not the default case)

markpollack:
Getting these test failure
```
java.lang.IllegalArgumentException: Can't find expected rendering result at BeanListTableModelTest-testExplicitPropertyNames.txt
	at org.springframework.util.Assert.notNull(Assert.java:112)
	at org.springframework.shell.table.AbstractTestWithSample.sample(AbstractTestWithSample.java:44)
	at org.springframework.shell.table.BeanListTableModelTest.testExplicitPropertyNames(BeanListTableModelTest.java:54)


java.lang.IllegalArgumentException: Can't find expected rendering result at BeanListTableModelTest-testSimpleConstructor.txt
	at org.springframework.util.Assert.notNull(Assert.java:112)
	at org.springframework.shell.table.AbstractTestWithSample.sample(AbstractTestWithSample.java:44)
	at org.springframework.shell.table.BeanListTableModelTest.testSimpleConstructor(BeanListTableModelTest.java:43)


java.lang.IllegalArgumentException: Can't find expected rendering result at BeanListTableModelTest-testHeaderRow.txt
	at org.springframework.util.Assert.notNull(Assert.java:112)
	at org.springframework.shell.table.AbstractTestWithSample.sample(AbstractTestWithSample.java:44)
	at org.springframework.shell.table.BeanListTableModelTest.testHeaderRow(BeanListTableModelTest.java:69)

```

markpollack:
Regarding the builder pattern, Table seems like the perfect candidate for such an approach.  The constructor, addBorder, format, align, wrap and size methods would typically be called once in the lifetime of the object, in order to create a specifically configured instance.  I doubt users of this class will call 'wrap' then render, then another all to wrap.  The steps to create this object are pretty complex, and in the example usage is also quite verbose, with `CellMatchers.ofType(Map.class)` being repeated over and over again, indicating that there could even be more room to clean up the creation of this object. So isolating the complex creation into a builder class leaving the Table class with a simple public API of only `render` and `getModel` is to me a much better solution.  While the Table class is also the builder, it seems cleaner to separate out the responsibilities.   If you aren't convinced, the class would look a bit more natural to me if all the method to configure the rendering etc, had the `add` prefix, much like `addBorder`, so `addFormatter`, `addAligner`, `addWrapper`, `addSizeConstraints`.

ericbottard:
@markpollack I added an embryo of `TableBuilder` in 20ad58d.
Let me know if that's ok with you (I did not want to waste time changing everything if the "feel" of it was wrong) and I can proceed. My intent would be to pass a read-only view of the LinkedHashMaps, that can be reversed there (thus short-circuiting the `getWrapper(row, col)` etc methods in Table)

ericbottard:
~~@markpollack Are you still getting those test failures you mentioned above? Builds fine for me.~~

Ah ah. turns out `./gradlew clean install` does **not** run the tests... I fixed those

markpollack:
Looks good.

ericbottard:
Updated the whole codebase with the new `TableBuilder` pattern. Nothing extraordinary since our last discussion, I'll merge the changes in the upcoming 24hrs unless you find a showstopper @markpollack, so that dependent PRs can move on.

markpollack:
I will take a look first thing this morning and merge

markpollack:
merged as 29e0d6ae7399633e29073c86e8a0a70b29efda00

PR #90, opened 2015-11-11
"help" behaves erratically if a command is also a prefix of another command

adityakishore:
Let's say if you have two commands "foo" and "foobar". Invoking "help foo" does not show detailed help of "foo" but minimal help for both "foo" and "foobar".

This pull request fixes this.


ericbottard:
Fixed via c8d2b7e which is simpler and does not require guava. The `help` command needs major revamping anyway, but thanks for this fix

PR #92, opened 2016-02-16
SHL-188: Add support for hidden CLI options

adityakishore:
This adds a new property `hidden` to `CliOption` annotation which will hide such options from `help`.

### DIFF
--- a/src/main/java/org/springframework/shell/core/SimpleParser.java
+++ b/src/main/java/org/springframework/shell/core/SimpleParser.java
@@ -1017,6 +1017,9 @@ public class SimpleParser implements Parser {
 						for (Annotation a : annotations) {
 							if (a instanceof CliOption) {
 								cliOption = (CliOption) a;
+								if (cliOption.hidden()) {
+									continue;
+								}
 
 								for (String key : cliOption.key()) {
 									if ("".equals(key)) {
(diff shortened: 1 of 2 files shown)


ericbottard:
Can you elaborate on the use case for this? Why would you want to hide *options* from the user. They're options after all, things that they can mess with, by definition.
Above all, even if there were different levels of discoverabilty for options, the only place where we'd **not** want to hide them is help, don't you think?

adityakishore:
May be "undocumented" is the right word here.

In cases where an advanced options for a user command, which is meant to be used by expert users only, it could be desirable to have such options hidden from the help.

ericbottard:
I'm really not in favor of having something not-documented. When you invoke a `man` page, you get everything, even the complex stuff (and indeed, even "experts" need to rely on documentation sometimes).
The help screen is very cumbersome anyway in its current (we plan on having a much better looking thing in Spring Shell 2, which is currently being prototyped.

adityakishore:
Sure.

adityakishore:
Regarding Spring Shell 2, is there a plan to move to SLF4J instead of JUL?

ericbottard:
Absolutely, the plan is to drop JUL, use jline3 (if it keeps moving), and have a complete pluggable approach. You can have a look at https://github.com/ericbottard/spring-shell2 (very, very rough edges)

PR #93, opened 2016-04-01
SHL-189 - Command classes can now be package protected.

olivergierke:
SimpleExecutionStrategy now makes sure the method to be invoked is accessible.


ericbottard:
Thanks @olivergierke !

Issue #96, opened 2016-05-06
Bootstrap starts StopWatch instance in main(), but stops in run()


Steve973:
If a user runs Bootstrap by invoking its main method, then everything works as expected.  But if I user creates an instance and manually invokes the run() method in their application, when run() finishes, it tries to stop the StopWatch instance, which throws an exception because it has never been started.  The StopWatch instance needs to be started at the beginning of the the run() method, so that stopping the stopwatch works as expected.  Pull request incoming for this issue.

PR #97, opened 2016-05-06
Moved sw.start() statement to beginning of run method.

Steve973:
Now the stopwatch is started and stopped in the same method, so no matter how the Bootstrap instance is run, the stopwatch instance will start and stop properly, and without an exception.


pdoranAtNovetta:
This fixes, #96 would be great to accept. +1

ericbottard:
Thanks for your contribution and original bug report. I went ahead and cleaned up `Bootstrap` even more in 1626a49

Issue #110, opened 2016-10-19
Allow cancel of currently input line using CTRL-C


ericbottard:
At the moment, pressing CTRL-C (which is a typing reflex used by many) simply triggers the shutdown of the shell.
It would be nice if one could simply cancel an already typed line, as many *nix shells do.
The proposed behavior is:

```
prompt>foo bar[CTRL-C]      // simply ignore what has been typed
prompt>[CTRL-C]             // if nothing has been typed, exit
```

See https://github.com/spring-projects/spring-shell/pull/108

Note that this is independent of killing a long running command execution, which may block the main prompt loop (which may or may not be addressed in the same PR for this)

PR #111, opened 2016-10-19
Allow cancel of entered line via CTRL-C

ericbottard:
Fixes https://github.com/spring-projects/spring-shell/issues/110


ericbottard:
Merged as fe512af

PR #112, opened 2016-11-03
SHL-146 Restore cursor position on flash

tim-patterson:
Fix for cursor reset instead of restore bug that was introduced in jline2 upgrade


ericbottard:
Thanks for catching this!

PR #123, opened 2017-04-30
[SHL-94] Support for interactive and masked inputs

camilojc:
Initial proposal to allow for interactive and masked inputs. Haven't used Spring Shell for too long so not fully familiar with the initial design, but tried to follow some concepts I saw whilst reading the code.

Any input more than welcome.

The main changes are:
- Added an ArgumentHolder (with 2 implementations) to allow for the ParseResult to hold parsing results that aren't yet fully resolved.
- Added an extra step in the AbstractShell to allow the implementations to resolve arguments that haven't been fully resolved (e.g. those that require interactive input)
- Included 2 extra flags in the CliOptions to allow for interactive and masked input values
- Split the Parser functionality to differentiate between the parsing of an argument value and the parsing of the command input.
- Moved the Parser classes to a separate parser sub-package

Main idea was to try and keep the AbstractShell and Parser agnostic from the choice of shell implementation to be used, whilst being able to allow for interactive input. 

Hope this helps.


camilojc:
Closed as discussed - to be implemented in v2 instead.

Issue #122, opened 2017-04-29
Package private command classes


anton-johansson:
#93 added support for having package private command classes. This works well for the `@CliCommand` annotation. However, it does not seem to work well for the `@CliAvailabilityIndicator` annotation. It never executes the annotated method and always seem to indicate `false`. The behavior changes when I change the class to `public`.

PR #124, opened 2017-05-02
Allow non public methods for @CliAvailabilityIndicator

ericbottard:
Fixes https://github.com/spring-projects/spring-shell/issues/122


ericbottard:
Merged as 685051b

PR #132, opened 2017-08-13
ParameterResolver extends Ordered

camilojc:
Hi @ericbottard - initial step in preparation for the other work as discussed ready for initial review

### DIFF
--- a/spring-shell-core/src/main/java/org/springframework/shell/ParameterResolver.java
+++ b/spring-shell-core/src/main/java/org/springframework/shell/ParameterResolver.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 the original author or authors.
+ * Copyright 2015-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -20,12 +20,21 @@ import java.util.List;
 import java.util.stream.Stream;
 
 import org.springframework.core.MethodParameter;
+import org.springframework.core.Ordered;
 
 /**
  * Implementations of this interface are responsible, once the command has been identified, of transforming the textual
  * input to an actual parameter object.
+ * 
+ * <p>
+ * An order can also be specified in case more than one {@link ParameterResolver} supports a {@link MethodParameter} via
+ * the {@link ParameterResolver#getOrder()} method.
+ * </p>
+ * 
+ * @author Eric Bottard
+ * @author Camilo Gonzalez
  */
-public interface ParameterResolver {
+public interface ParameterResolver extends Ordered {
 
 	/**
 	 * Should return true if this resolver recognizes the given method parameter (<em>e.g.</em> it
--- a/spring-shell-core/src/main/java/org/springframework/shell/ParameterResolver.java
+++ b/spring-shell-core/src/main/java/org/springframework/shell/ParameterResolver.java
@@ -20,21 +20,21 @@ import java.util.List;
 import java.util.stream.Stream;
 
 import org.springframework.core.MethodParameter;
-import org.springframework.core.Ordered;
+import org.springframework.core.annotation.AnnotationAwareOrderComparator;
 
 /**
  * Implementations of this interface are responsible, once the command has been identified, of transforming the textual
  * input to an actual parameter object.
  * 
  * <p>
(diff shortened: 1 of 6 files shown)


spring-shell-shell1-adapter/src/main/java/org/springframework/shell/legacy/LegacyParameterResolver.java
```diff
@@ -192,4 +192,8 @@ public ParseResult(String value, Integer from) {
 		
 	}
 	
+	@Override
```
idrissov: what if use default method in interface?

spring-shell-core/src/main/java/org/springframework/shell/ParameterResolver.java
```diff
@@ -20,12 +20,21 @@
...
+ * <p>
+ * An order can also be specified in case more than one {@link ParameterResolver} supports a {@link MethodParameter} via
+ * the {@link ParameterResolver#getOrder()} method.
+ * </p>
+ * 
+ * @author Eric Bottard
+ * @author Camilo Gonzalez
  */
-public interface ParameterResolver {
+public interface ParameterResolver extends Ordered {
```
ericbottard: I wouldn't force `Ordered` on all resolvers, but simply sort them using `AnnotationAwareOrderComparator.sort()` after injection

spring-shell-core/src/main/java/org/springframework/shell/Shell.java
```diff
@@ -264,4 +284,24 @@ private String findLongestCommand(String prefix) {
 		return "".equals(result) ? null : result;
 	}
 
+	private static class ParameterToResolver implements Comparable<ParameterToResolver> {
```
ericbottard: I don't understand the need for that new tuple class. Why not just sort the resolvers (see my comment above) and continue to return the first that matches? Why put the parameter index in the picture? There is something I'm not understanding here, let's discuss tomorrow on the call if that helps

camilojc: Let's assume we have 2 parameters, e.g. `methodName(T1 arg1, T2 arg2)`, and 3 parameter resolvers `[P1, P2 and P3]` (in priority order) and the possible resolvers for `arg1` -> `P1` and `P3`, and for `arg2` -> `P2` and `P3`. During a `resolve` operation, my understanding is that the operation should attempt the following in order:

1. try to resolve `arg1` using `P1`
2. try to resolve `arg2` using `P2`
3. try to resolve `arg1` using `P3` (in case `P1` failed to resolve)
4. try to resolve `arg2` using `P3` (in case `P2` failed to resolve)

In 1 and 3 are why we need to return multiple resolvers (and not the first that matches) as we'd need to try the next compatible resolver (respecting the priority of the other resolvers).

In 3 and 4 are why the index comes into place, the same resolver should be invoked in parameter order.

The reason for the tuple is to pre-compute the order the operation will use.

In a more concrete example, the way I see this is what we discussed last time as a possible approach, for example, StandardParameterResolver being potentially broken down into:

1. `P1` -> StandardNamedParameterResolver
2. `P2` -> StandardBooleanParameterResolver (I don't actually had this in mind, but works with the example above :) - but the boolean 0-arity case could actually be handled like this, again, just for the example, but worth thinking about)
3. `P3` -> StandardPositionalParameterResolver

Let me know if you had a different idea in mind, let's try to discuss tomorrow, but will let you know if I need to cancel (in case of family news :) )

spring-shell-core/src/main/java/org/springframework/shell/Shell.java
```diff
@@ -185,10 +190,17 @@ private boolean noInput(Input input) {
 			// Try to complete arguments
 			MethodTarget methodTarget = methodTargets.get(best);
 			Method method = methodTarget.getMethod();
-			Arrays.stream(method.getParameters())
-				.map(Utils::createMethodParameter)
-				.flatMap(mp -> findResolver(mp).complete(mp, argsContext).stream())
-				.forEach(candidates::add);
+			
+			List<MethodParameter> parameters = Utils.createMethodParameters(method).collect(Collectors.toList());
+			validateParameters(parameters);
```
ericbottard: I think we can move this validation at startup time when we discover MethodTargets.
Will do on merge

ericbottard:
Merged as 6c231a0 with slight rework. Eager validation at start instead of each usage

PR #161, opened 2017-09-23
Add optional command grouping

### LINKED ISSUE #135: Add optional command grouping
_From @ericbottard on June 2, 2017 8:24_

Add an optional category to commands, to be used e.g. in `help` screen

_Copied from original issue: ericbottard/spring-shell2#82_

rweisleder:
The output of `help` command is now grouped. Commands without group are in a group called "Default" on top.

Resolves #135

### DIFF
--- a/spring-shell-core/src/main/java/org/springframework/shell/MethodTarget.java
+++ b/spring-shell-core/src/main/java/org/springframework/shell/MethodTarget.java
@@ -37,16 +37,22 @@ public class MethodTarget {
 
 	private final String help;
 
+	private final String group;
+
 	/**
 	 * If not null, returns whether or not the command is currently available. Implementations must be idempotent.
 	 */
 	private final Supplier<Availability> availabilityIndicator;
 
 	public MethodTarget(Method method, Object bean, String help) {
-		this(method, bean, help, null);
+		this(method, bean, help, null, null);
 	}
 
 	public MethodTarget(Method method, Object bean, String help, Supplier<Availability> availabilityIndicator) {
+		this(method, bean, help, null, availabilityIndicator);
+	}
+
+	public MethodTarget(Method method, Object bean, String help, String group, Supplier<Availability> availabilityIndicator) {
 		Assert.notNull(method, "Method cannot be null");
 		Assert.notNull(bean, "Bean cannot be null");
 		Assert.hasText(help, String.format("Help cannot be blank when trying to define command based on '%s'", method));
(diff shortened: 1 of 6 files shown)


pivotal-issuemaster:
@rweisleder Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-shell&pullRequestId=161)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-shell&pullRequestId=161) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@rweisleder Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-shell&pullRequestId=161)!

ericbottard:
Thanks for this awesome contribution!
I think I'll extract description + group in a dedicated object, but can do later after merging this PR (almost) as is.

Here are a couple of ideas, let me know what you think about them as a user:
- why not turn the class name where a command is defined into a group (turning CamelCase into separate words) by default?
- `@ShellComponent` could have a `group` attribute to customize the name, and even a special value to turn that automatic behavior off at the class level if desired

rweisleder:
I like more the idea how Gradle handles task grouping: If you not specify a group, all tasks are in a group called "Other tasks". (opt-in)

To change your idea to opt-in: What about a new class annotation `@ShellMethodGroup` to specify a group for all commands in a class? One could also separate the commands into two classes but have them in one group just by adding the annotation to both classes.
Currently one have to copy and paste the group parameter for each method.

ericbottard:
There are two orthogonal aspects:
- opt-in vs. opt-out
- mass-specifying a group at some higher level than the command (typically, the class)

For the opt-in/opt-out debate, I am more in favor of opt-out, because it is more in line with the Boot philosophy that users are accustomed to, and because the Class is already a very sensible group indicator for commands.

For the second concern, I think we're basically saying the same thing, with the devil in the details of whether to use a different annotation or not (I'd rather not)

rweisleder:
So in summary I think this code
```
@ShellComponent
public class FirstComponentGroup {

    @ShellMethod
    public void methodOne() {...}

    @ShellMethod(group = "Second group")
    public void methodTwo() {...}
}

@ShellComponent(group = "Second group")
public class SecondComponentGroup {

    @ShellMethod
    public void otherMethod() {...}
}
```
should lead to this command grouping
```
First component group
    method-one

Second group
    method-two
    other-method
```
which sounds good to me.

ericbottard:
You got that all right (still thinking about Title Case for groups or not)

ericbottard:
Created https://github.com/spring-projects/spring-shell/issues/163

PR #189, opened 2017-12-24
Support execute-then-quit as arguments in addition to @/path/to/file

### LINKED ISSUE #183: Remove component scan
As scan is evil, remove `@ComponentScan` from `SpringShellAutoConfiguration` in favour of defining beans manually. This allows customisation of commands with conditionals(assuming user knows what to override).

### LINKED ISSUE #187: Make shell understand non-interactive mode
In skipper we wanted to run commands as non-interactive mode
```
java -jar spring-cloud-skipper-shell/target/spring-cloud-skipper-shell-1.0.0.BUILD-SNAPSHOT.jar search

vs.

java -jar spring-cloud-skipper-shell/target/spring-cloud-skipper-shell-1.0.0.BUILD-SNAPSHOT.jar
skipper:>search
```
We found a quick hack to this in [InteractiveModeApplicationRunner](https://github.com/spring-cloud/spring-cloud-skipper/blob/master/spring-cloud-skipper-shell-commands/src/main/java/org/springframework/cloud/skipper/shell/command/support/InteractiveModeApplicationRunner.java#L99).

What we also wanted to do is to return non zero from an non-interactive mode when underlying shell command fails. This turned out to be impossible, afaik, to implement atop of shell core in a skipper. There is `ExitRequestResultHandler` and `ExitRequest` but I wasn't able to use those without breaking interactive mode. I thought by customising shell, created #182 but that was also dead end.

I think we need to make shell core to better understand non-interactive run.

ericbottard:
Make ResultHandlers configuration more explicit 
Handle exit as a dedicated case (prevents eg 'exit' commands in scripts to make script quit) 
Add an example of custom ApplicationRunner

Fixes #187
Fixes #183


ericbottard:
Merry Christmas @jvalkeal 
Have a look at this PR, I think it could get you what you want to achieve even without allowing override of specific ResultHandlers. Try:

```
java -jar ./spring-shell-samples/target/spring-shell-samples-2.0.0.BUILD-SNAPSHOT.jar help
```

and
```
java -jar ./spring-shell-samples/target/spring-shell-samples-2.0.0.BUILD-SNAPSHOT.jar fail --element-type TYPE
```

jvalkeal:
This seem to have a side-effect for `exit` not exiting from interactive mode properly. Actually you need to exit twice.

Current master:
```
$ java -jar ./spring-shell-samples/target/spring-shell-samples-2.0.0.BUILD-SNAPSHOT.jar
my-shell:>exit
$
```
This PR:
```
$ java -jar ./spring-shell-samples/target/spring-shell-samples-2.0.0.BUILD-SNAPSHOT.jar
my-shell:>exit
shell:>exit
$
```

ericbottard:
This is weird. I don't get the behavior you describe, but I do get something odd:
my prompt is not being customized to `my-shell:>`, while it should

jvalkeal:
It now works slightly better but looks like you get into trouble with options:
```
✔ ~/repos/spring/spring-shell [ericbottard-refactor-result-handlers L|✔] 
11:14 $ java -jar ./spring-shell-samples/target/spring-shell-samples-2.0.0.BUILD-SNAPSHOT.jar

  .   ____          _            __ _ _
 /\\ / ___'_ __ _ _(_)_ __  __ _ \ \ \ \
( ( )\___ | '_ | '_| | '_ \/ _` | \ \ \ \
 \\/  ___)| |_)| | | | | || (_| |  ) ) ) )
  '  |____| .__|_| |_|_| |_\__, | / / / /
 =========|_|==============|___/=/_/_/_/
 :: Spring Boot ::        (v1.5.8.RELEASE)

my-shell:>sum --v1 1 --v2 2
3
my-shell:>exit
✔ ~/repos/spring/spring-shell [ericbottard-refactor-result-handlers L|✔] 
11:14 $ java -jar ./spring-shell-samples/target/spring-shell-samples-2.0.0.BUILD-SNAPSHOT.jar sum --v1 1 --v2 2

  .   ____          _            __ _ _
 /\\ / ___'_ __ _ _(_)_ __  __ _ \ \ \ \
( ( )\___ | '_ | '_| | '_ \/ _` | \ \ \ \
 \\/  ___)| |_)| | | | | || (_| |  ) ) ) )
  '  |____| .__|_| |_|_| |_\__, | / / / /
 =========|_|==============|___/=/_/_/_/
 :: Spring Boot ::        (v1.5.8.RELEASE)

Anonymous option has already been set
✘-1 ~/repos/spring/spring-shell [ericbottard-refactor-result-handlers L|✔] 

```
I'm not sure if this should now work OOB or if arguments needs some sort of a custom handling.

ericbottard:
This is because I used an `ApplicationRunner` and happened to test with the one command that doesn't use `--foo` (`help -C <command>`).

I changed the example to use a `CommandLineRunner`, which you'll need to do as well in skipper. Obviously, if you want to handle some `--foo` parameters on the command line, you'll need to resolve the ambiguity.
Another approach (but with added quoting dilemmas) is to parse the command(s) you want to execute-then-quit like this:
```
skipper-shell 'command --arg1 value1 --arg2 value2' 'another-command --arg value'
```

ericbottard:
Merged as f90edd42c41c738ed628805922c850c9b05c6730

jeremymailen:
Hi Eric,

Any reason not to include a CommandLineApplicationRunner as standard instead of just the example? I can't think of too many downside since the `shellapp @some/file/path` can simultaneously be supported.

In any case spring-shell 1.x had that behavior and we use it quite a bit for automation.

ericbottard:
The principal reason is that Spring Shell 2 applications are typically Spring Boot applications, and as such may certainly use the `--foo.bar=x` mechanism to configure their environment. So this would collide and provide unexpected results

ccaspanello:
@ericbottard - Can you give me a few details on how the "non-interactive" mode functionality should work.

Sample code works great in interactive mode;
```
k2:>add 1 2
add(a: 1, b: 2)    <--- from LOG.info() line
3                         <--- from return statement of a+b
```

However, after I add **spring.shell.interactive.enabled=false** to **application.properties** the console exits right away.  It doesn't look like the command is running.  I tried running the help command with the syntax mentioned above and that didn't even work.  Any help would be greatful.  FYI - I did clone the repo and build the code to make sure I had the latest and greatest.

ericbottard:
@ccaspanello The feature described here is not available "out-of-the-box". The setting you mention is *supporting* it, but additional setup is required. Have a look at [ExampleApplicationRunnerConfiguration](https://github.com/spring-projects/spring-shell/blob/c3081d260e7f797f3f2ccd77aedd7decc385d536/spring-shell-samples/src/main/java/org/springframework/shell/samples/ExampleApplicationRunnerConfiguration.java)

PR #197, opened 2018-01-17
Fix method annotation in some reference documentation snippets

fwaibel:
Somehow `@ShellCommand` found its way into the snippets. This commit
replaces it with the correct annotation `@ShellMethod`.

Keep up the good work!

PS: I stumbled upon this typo while writing "First hand experience migrating from Spring Shell 1.2 to 2.0" -  https://devops.datenkollektiv.de/migrating-from-spring-shell-12-to-20.html today.

PR #198, opened 2018-01-17
Fix version of parent pom in spring-shell-samples

fwaibel:
Travis rejected my pull request https://github.com/spring-projects/spring-shell/pull/197 due to a wrong version of the parent pom in spring-shell-samples. Since this seem unrelated I created another pull request.


ericbottard:
Again, nice catch. I've noticed this as well (comes from the fact that `samples` are not published and hence not picked up by our release process...) and will be part of another PR already open.

PR #234, opened 2018-10-03
try to avoid reflective access that make Java 9+ unhappy

joshlong:
I want to run Spring Shell on Java 9+. I'm using Java 11. It barks when we reflectively try to change the visibility of fields/methods on classes in the JDK.


ericbottard:
Merged as ffcebbf

PR #246, opened 2019-01-15
Update JCommander version

Timunas:
This commit updates the JCommander version to 1.72.
The update also takes into account an issue that was introduced in new versions of the JCommander, namely, https://github.com/cbeust/jcommander/issues/377.


spring-shell-jcommander-adapter/src/main/java/org/springframework/shell/jcommander/JCommanderParameterResolver.java
```diff
@@ -104,9 +104,7 @@ public ValueResult resolve(MethodParameter methodParameter, List<String> words)
 
 	private JCommander createJCommander(MethodParameter methodParameter) {
 		Object pojo = BeanUtils.instantiateClass(methodParameter.getParameterType());
-		JCommander jCommander = new JCommander(pojo);
-		jCommander.setAcceptUnknownOptions(true);
```
Timunas: This was removed because of the issue described here cbeust/jcommander#377.

jvalkeal:
Thx, I'm merging this to main and picking 1.81.

jvalkeal:
Merged per 63f786fd8c84118c789af139389a30ab22dffd56

PR #248, opened 2019-01-21
Download maven-wrapper.jar if it is absent instead of keeping it in repository

snuyanzin:
The PR provides possibility not to keep maven-wrapper.jar in the repository but download it in case it is absent, also it adds maven-wrapper.jar into .gitignore
Also in case of Windows it uses TLS1.2 (as by default Windows does not use it while GitHub requires TLS1.2)


pivotal-issuemaster:
@snuyanzin Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-shell&pullRequestId=248)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-shell&pullRequestId=248) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@snuyanzin Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-shell&pullRequestId=248)!

jvalkeal:
I'm going to reject this PR. While scripts we have right now already supports automatic download we want to keep it as is for now as other spring projects use it like this. Also start.spring.io adds wrapper jar. If/when these change in spring landscape we follow the light.

PR #275, opened 2019-03-28
URL Cleanup

spring-operator:
- Ensure Gradle Wrapper is downloaded via https
- This project uses an old version of Gradle in which mavenCentral() and
  jcenter() use http instead of https. This commit switches to use an
  explicit URL so https is used.


jvalkeal:
stale

Issue #212, opened 2018-04-02
Support for Optional<> parameters


jasonwjones:
I have several commands where a parameter is optional (in the sense that it has a `defaultValue = ShellOption.NULL` annotation) but there isn't a default value that makes sense to assign statically. Based on how the code works in several of these methods, the way I can treat these is by wrapping them in `Optional` or adding logic to assign them an appropriate default value at runtime.

It seems that it might be possible to just implement a `Converter<String, Optional<Foo>>` but given that `Optional` is a foundational type (and Spring Shell relies on Java 1.8+), I think it would be nice to support `Optional` "out of the box" for method parameter injection, in conjunction with the current support for Converter.

For example, I may have this method:

```
public void comments(@ShellOption Foo foo) { ... }
```

And already have a corresponding converter:

```
@Component
public class FooConverter implements Converter<String, Foo> { ... }
```

Given this it'd be nice if this worked, with no additional converters needed:

```
public void comments(@ShellOption Optional<Foo> foo) { ... }
```

Thoughts?

c4deszes:
Made a pull request #287 

I support this feature and actually the current resolver will fill `Optional<T>` but if the parameter doesn't exist it will fail. The solution is to specify `@ShellOption.optOut` and have a custom ParameterResolver that supports Optional and resolves everything into `Optional.empty()`. But same story as with `defaultValue` because you would have to specify optOut for every Optional argument.

jvalkeal:
We're doing housekeeping in issues and closing ones which are potentially stale due to recent changes in a project. We are now more up to date with `Spring Boot` starting with release `2.1.0` and work in `3.x` making shell compatible with `Spring Framework 6` and `Spring Boot 3`.

If you think issue should not have been closed, try to re-open or open a new issue with updated info.

PR #287, opened 2019-10-07
Provides support for Optional<>, @Nullable

c4deszes:
Solves #212 

Currently `StandardParameterResolver` is responsible for converting command arguments into objects, if an argument is missing then it can use a default value if something other than `"__NONE__"` is used, if `"__NULL__ "` is used then the `ParameterResolver` will inject a null value for that parameter.

This makes optional arguments quite annoying to implement as you have to annotate every argument with `@ShellOption(defaultValue = ShellOption.NULL)` then check it's value in the method.

`StandardParameterResolver` uses a `ConversionService` to convert Strings into Objects, and there's actually a default converter that can do T -> Optional<T>. But the current behavior is that if the parameter is missing it throws a `RuntimeException`, which clearly isn't the intention when using `Optional<T>` as a parameter.

The way you can avoid `StandardParameterResolver` is by setting `@ShellOption.optOut` to `true`, then you could specify your custom one that supports Optionals, but then once again you would have to annotate every parameter.

So there are three ways we could support this functionality:
1. Change the `supports` function of `StandardParameterResolver` to only include parameters it can actually resolve, right now it claims to support everything unless you opt out.
2. Change the `resolve` function of `StandardParameterResolver` to return `Optional.empty()` if the parameter is missing and the parameter is of `Optional` type
3. Fallback to alternative `ParameterResolver`s in case of a `RuntimeException`, if all of them fail we could throw the last exception or allow `validateArgs` in `Shell` to do the work

This commit allows the usage of `Optional<T>` and `@Nullable `annotation without modifying the `StandardParameterResolver` (Option 3). Rather it changes the way the Shell chooses `ParameterResolver`s, this ensures compatibility with existing code, most users likely never used custom `ParameterResolver`s instead used `ValueProvider`s and `Converter`s.

The new behavior is that in case of a `ParameterMissingResolutionException` the `Shell` will try other resolvers. If the parameter is there then it will be resolved by the `StandardParameterResolver` which is now annotated with `@Order(1)` so it's going to be the first resolver being tried.

The added `NullableParameterResolver` and `OptionalParameterResolver` will resolve arguments into `null` and `Optional.empty()` respectively. With `@Nullable` tab completion is supported, but with Optionals this functionality is non existent right now.

Flagged `@ShellOption.optOut` as `@Deprecated` because the new behavior allows the usage of custom `ParameterResolver`s without having to opt out.

Examples:
```
@ShellMethod(value = "Greets users")
public String greet(Optional<String> name) {
    if(name.isPresent()) {
        return "Hello, " + name.get() + "!";
    }
    return "Hello, World!";
}
```

```
@ShellMethod(value = "Greets users")
public String greet(@Nullable String name) {
    if(name != null) {
        return "Hello, " + name + "!";
    }
    return "Hello, World!";
}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,6 @@ target/
 *.iws
 .DS_Store
 spring-shell.log
+
+.vscode/
+.sts4-cache/
--- a/pom.xml
+++ b/pom.xml
@@ -11,7 +11,7 @@
 	<parent>
 		<groupId>org.springframework.boot</groupId>
 		<artifactId>spring-boot-starter-parent</artifactId>
-		<version>2.0.6.RELEASE</version>
+		<version>2.1.9.RELEASE</version>
 		<relativePath />
 	</parent>
 	<scm>
@@ -35,6 +35,7 @@
 		<module>spring-shell-standard-commands</module>
 		<module>spring-shell-jcommander-adapter</module>
 		<module>spring-shell-table</module>
+		<module>spring-shell-websocket</module>
 		<module>spring-shell-docs</module>
 		<module>spring-shell-dependencies</module>
 		<module>spring-shell-starter</module>
@@ -77,6 +78,11 @@
 				<groupId>org.springframework.shell</groupId>
 				<artifactId>spring-shell-table</artifactId>
 				<version>3.0.0.BUILD-SNAPSHOT</version>
+			</dependency>
+			<dependency>
+				<groupId>org.springframework.shell</groupId>
+				<artifactId>spring-shell-websocket</artifactId>
+				<version>1.0.0.BUILD-RELEASE</version>
 			</dependency>
 			<!-- OTHER DEPS -->
 			<dependency>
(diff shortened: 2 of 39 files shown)


c4deszes:
I think the current opt out behavior is quite a limitation and it makes custom `ParameterResolver`s inconvenient.

Some new features it would allow:
- Grabbing variables outside of the Shell context, for example we could have Principal parameters autofilled from the `SecurityContextHolder`
- Simple variable argument size support, an example [Spring Shell System call](https://github.com/c4deszes/spring-snippets/blob/master/shell-system-commands/src/main/java/com/example/demo/ShellCommands.java) This would be more functional if I didn't have to specify the `optOut` parameter

In my opinion a better solution would be to have `@ShellOption` have a `resolver` field, defaulting to null. If it's null then it uses the fallback method brought by this pull request, otherwise it would force that `ParameterResolver`.

Issue #235, opened 2018-10-08
Need to call operating system commands in the spring shell


zengqingjun:
Need to call operating system commands in the spring shell。
How to use？

ericbottard:
Until infinite arity for parameters, this is gonna be a little awkward but:

Create a Spring Shell command that does that (you may want to call it `!` for example). Such a command would read the rest of its arguments (either you hardcode the number of args because you know which system commands you want to support, or use just one arg which you'll split yourself on spaces -- the use will have to surround it with quotes) and turn that into the command to execute (using JDK's `Process` class).

jvalkeal:
We're doing housekeeping in issues and closing ones which are potentially stale due to recent changes in a project. We are now more up to date with `Spring Boot` starting with release `2.1.0` and work in `3.x` making shell compatible with `Spring Framework 6` and `Spring Boot 3`.

If you think issue should not have been closed, try to re-open or open a new issue with updated info.

Issue #140, opened 2017-08-16
Implement infinite arity


ericbottard:
_From @ericbottard on January 7, 2016 11:56_

_Copied from original issue: ericbottard/spring-shell2#11_

ericbottard:
Blocks https://github.com/spring-cloud/spring-cloud-skipper/issues/316

Coada:
+1 for this feature. I need it to properly map it to varargs.

jvalkeal:
We're doing housekeeping in issues and closing ones which are potentially stale due to recent changes in a project. We are now more up to date with `Spring Boot` starting with release `2.1.0` and work in `3.x` making shell compatible with `Spring Framework 6` and `Spring Boot 3`.

If you think issue should not have been closed, try to re-open or open a new issue with updated info.

PR #312, opened 2020-11-08
Clarify how to achieve varying arity in docs

JohnnyMarnell:
Add examples of how to achieve, with existing functionality, a varying amount of parameter arity (and favor this verbage over "infinite", which, though technically true, could be misleading for evoking neding a very large amount of a parameter, instead of the normal usecase of a few of any amount).

The docs have long said "TO BE IMPLEMENTED", however this functionality is already achievable. I needed this (common IMO) use case and searched in the docs, plus found some long standing open issues like [here](https://github.com/spring-projects/spring-shell/issues/140) and [here](https://github.com/spring-projects/spring-shell/issues/235), but was happy to find I could get it to work without needing any new features.

### DIFF
--- a/spring-shell-docs/src/main/asciidoc/using-spring-shell.adoc
+++ b/spring-shell-docs/src/main/asciidoc/using-spring-shell.adoc
@@ -300,8 +300,27 @@ shell:>add --numbers 1 --numbers 2 --numbers 3.3
 ----
 ====
 
-===== Infinite Arity
-TO BE IMPLEMENTED
+===== Varying Amount Arity
+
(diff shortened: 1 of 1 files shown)


JohnnyMarnell:
@ericbottard Hello, do you or someone you can connect me with, have any thoughts / suggestions here?

iuliu-b:
Hi - what's the status of this PR? The docs [are still not updated](https://docs.spring.io/spring-shell/docs/current-SNAPSHOT/reference/htmlsingle/#_infinite_arity), and this is super useful

JohnnyMarnell:
🦗's... 🙏🏻 @jvalkeal ?

jvalkeal:
Thx, sorry for delay. Merged per 2b7a9b15e64df16ba1ee6f793e43637ebc48b0f7.

PR #320, opened 2021-05-18
Update document appearance

Buzzardo:
Use Spring's standard styles for the ToC. Also added epub generation.


jvalkeal:
Merged per 012481ef63fc3bb5c7ec09afff0b9f7fd1cc0e98. Had to do a bit of fixes as it was old PR.

PR #355, opened 2022-01-24
Make grouping configurable in build-in help

### LINKED ISSUE #347: Change or remove group of "Built-in Commands"
When the user types "help", the Built-In Commands are always shown as a separate group, followed by a group of my own commands (named after the class they are in).

I would like to see all the commands together, with no distinction.  But there doesn't appear to be a way to "ungroup" all the commands, or alternatively merge the groups together.

Can we get some more flexibility on the Built-In commands (short of having to redefine every one of them)?

jvalkeal:
- New option spring.shell.command.help.grouping-mode which can be either flat/group.
- Build-in help command can now choose if to group command or just show flat list.
- Fixes #347

### DIFF
--- a/spring-shell-autoconfigure/src/main/java/org/springframework/shell/boot/SpringShellProperties.java
+++ b/spring-shell-autoconfigure/src/main/java/org/springframework/shell/boot/SpringShellProperties.java
@@ -126,6 +126,7 @@ public class SpringShellProperties {
 	public static class HelpCommand {
 
 		private boolean enabled = true;
+		private GroupingMode groupingMode = GroupingMode.GROUP;
 
 		public boolean isEnabled() {
 			return enabled;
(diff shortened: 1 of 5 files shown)


jvalkeal:
This would result for example:
```
AVAILABLE COMMANDS

  completion bash: Generate bash completion script
  help: Display help about available commands.
  history: Display or save the history of previously run commands
  script: Read and execute commands from a file.
  version: Show version info
  dependencies: List supported dependencies
  info: Show the Initializr server being used
  init: Initialize project
```
vs.
```
AVAILABLE COMMANDS

Built-In Commands
        completion bash: Generate bash completion script
        help: Display help about available commands.
        history: Display or save the history of previously run commands
        script: Read and execute commands from a file.
        version: Show version info

Initializr Commands
        dependencies: List supported dependencies
        info: Show the Initializr server being used
        init: Initialize project
```

PR #358, opened 2022-01-31
Add non-interactive shell runner customizer

onobc:
- Add ordering on the ApplicationShellRunner which is different than the order on the DefaultApplicationRunner. The latter is order that the default runner will execute wrt other application runners. However, the shell runner order is the order that the default app runner will consider them for "1st available to handle the shell".

- Refactored the shell runners javadocs to clearly spell out the above and make them consistent w/ the new mechanism (not the enabled property of days old). 

- Added customizer which allows us to replace the skipper InteractiveApplicationRunner w/ a customizer that simply filters out the non-command args from the `ApplicationArguments` (@jvalkeal )

~This is WIP as I have not added any unit/int testing. I have manually verified the interactive and non-interactive cases in skipper manually though. (https://github.com/spring-cloud/spring-cloud-skipper/pull/1026)~

### DIFF
--- a/spring-shell-core/src/main/java/org/springframework/shell/DefaultApplicationRunner.java
+++ b/spring-shell-core/src/main/java/org/springframework/shell/DefaultApplicationRunner.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 the original author or authors.
+ * Copyright 2021-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


spring-shell-core/src/main/java/org/springframework/shell/DefaultApplicationRunner.java
```diff
@@ -23,17 +23,24 @@
...
- * Default {@link ApplicationRunner} which dispatches to first ordered
- * {@link ShellRunner} able to handle shell.
+ * Default {@link ShellApplicationRunner} which dispatches to the first ordered {@link ShellRunner} able to handle
+ * the shell.
  *
  * @author Janne Valkealahti
+ * @author Chris Bono
  */
+@Order(DefaultApplicationRunner.PRECEDENCE)
 public class DefaultApplicationRunner implements ShellApplicationRunner {
```
onobc, line 37: **[UNRELATED]** I would like to rename this to `DefaultShellApplicationRunner` for specificity.

jvalkeal: Yeah true, we could do it with this pr.

jvalkeal:
Other than that, I think this looks good!

PR #371, opened 2022-02-18
Simplify some expressions, typo fixes

ullenboom:
Pls reformat ...


onobc:
Hi @ullenboom 

While I do agree w/ some of the simplification and improvements, we do not typically consider PRs that are widespread format/style/improvements like this. 

1. It is a large PR - it is hard to find the time to go through this many files (even if they seem small and low-risk, someone has to review). 
2. The style/formatting added does not follow the format prescribed by the other Spring/Cloud projects (eg extra spaces before parens is not something you will find). 
3. Final is typically only used for immutable value objects (this does fit w/ some of your changes in this area though). 

The recommended approach is to do clean up like this as we go, when we are in that area of code for some functional change/fix. 

Another possible route would be to create a very targeted proposal that does 1 of the things on your list, such as "Make constructor injected configuration class members final to enforce immutablility". A targeted proposal is more likely to be accepted.

I do also realize this project has close to zero contributing guidelines, nor automated checkstyle/format rules. For that, I created #373 .

PR #372, opened 2022-02-18
Add support for multiple non-interactive commands

onobc:
Also:
* Rename ThemeResolver bean to avoid clash w/ Spring Boot registered ThemeResolver
* Move shell runner precedence to public static field to allow extension/access

This was needed to support `spring.shell.commandFile` option in SCDF shell. I originally implemented it in SCDF by creating a separate NonInteractiveShellRunner. However, its cleaner to simply support N commands here and let the `commandfile->commands` come in via customizer in SCDF (see [here](https://github.com/spring-cloud/spring-cloud-dataflow/pull/4835/files#diff-7c1674166952308591746a47ec386ac05f55151228991c818a825dfe74b59e07R74)).


jvalkeal:
thx, lgtm

PR #381, opened 2022-02-26
Add dynamic command registration

### LINKED ISSUE #379: Register commands dynamically
It'd be nice if we could handle command registrations dynamically when shell is running. We could start with something simple like allow add/remove commands from a `CommandRegistry`.

jvalkeal:
- Extend CommandRegistry for add/remove methods.
- For rest of shell classes move to use registry
  directly instead of caching commands as registry
  is not immutable anymore.
- Add new sample
- Fixes #379

### DIFF
--- a/spring-shell-core/src/main/java/org/springframework/shell/ConfigurableCommandRegistry.java
+++ b/spring-shell-core/src/main/java/org/springframework/shell/ConfigurableCommandRegistry.java
@@ -58,6 +58,16 @@ public class ConfigurableCommandRegistry implements CommandRegistry {
 				.collect(Collectors.toMap(e -> e.getKey(), e -> e.getValue()));
 	}
 
+	@Override
+	public void addCommand(String name, MethodTarget target) {
+		commands.put(name, target);
+	}
+
+	@Override
+	public void removeCommand(String name) {
+		commands.remove(name);
+	}
+
 	public void register(String name, MethodTarget target) {
 		MethodTarget previous = commands.get(name);
 		if (previous != null) {
--- a/spring-shell-core/src/main/java/org/springframework/shell/MethodTarget.java
+++ b/spring-shell-core/src/main/java/org/springframework/shell/MethodTarget.java
@@ -72,6 +72,14 @@ public class MethodTarget implements Command {
 	 * Construct a MethodTarget for the unique method named {@literal name} on the given object. Fails with an exception
 	 * in case of overloaded method.
 	 */
+	public static MethodTarget of(String name, Object bean, String description, String group) {
+		return of(name, bean, new Help(description, group));
+	}
+
+	/**
+	 * Construct a MethodTarget for the unique method named {@literal name} on the given object. Fails with an exception
+	 * in case of overloaded method.
+	 */
 	public static MethodTarget of(String name, Object bean, Help help) {
 		return of(name, bean, help, null);
 	}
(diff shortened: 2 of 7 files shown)


spring-shell-core/src/main/java/org/springframework/shell/ConfigurableCommandRegistry.java
```diff
@@ -58,6 +58,16 @@ else if (mim == InteractionMode.NONINTERACTIVE) {
 				.collect(Collectors.toMap(e -> e.getKey(), e -> e.getValue()));
 	}
 
+	@Override
+	public void addCommand(String name, MethodTarget target) {
+		commands.put(name, target);
```
onobc, line 63: Should we log warning when a command is overrwrritten? Not sure when this would happen, but it may be helpful to know when/if it does.

spring-shell-core/src/main/java/org/springframework/shell/MethodTarget.java
```diff
@@ -68,6 +68,14 @@ public MethodTarget(Method method, Object bean, Help help, Supplier<Availability
 		this.interactionMode = interactionMode;
 	}
 
+	/**
+	 * Construct a MethodTarget for the unique method named {@literal name} on the given object. Fails with an exception
+	 * in case of overloaded method.
+	 */
```
onobc, line 74: Do you mind adding doc for the params?

spring-shell-core/src/main/java/org/springframework/shell/Shell.java
```diff
@@ -158,6 +143,7 @@ public Object evaluate(Input input) {
 
 		List<String> words = input.words();
 		if (command != null) {
+			Map<String, MethodTarget> methodTargets = commandRegistry.listCommands();
```
onobc, line 146: A slight improvement would be to move this into a `methodTargets()` method (used in 3 places) and it may also be an easy area of extension to allow getting commands in a slightly different fashion.

spring-shell-samples/src/main/java/org/springframework/shell/samples/standard/RegisterCommands.java
```diff
@@ -0,0 +1,65 @@
...
+import org.springframework.shell.MethodTarget;
+import org.springframework.shell.standard.AbstractShellComponent;
+import org.springframework.shell.standard.ShellComponent;
+import org.springframework.shell.standard.ShellMethod;
+import org.springframework.shell.standard.ShellOption;
+
+@ShellComponent
+public class RegisterCommands extends AbstractShellComponent {
+
+	private final PojoMethods pojoMethods = new PojoMethods();
```
onobc, line 27: **[NIT]** Indention

spring-shell-samples/src/main/java/org/springframework/shell/samples/standard/RegisterCommands.java
```diff
@@ -0,0 +1,65 @@
...
+package org.springframework.shell.samples.standard;
+
+import org.springframework.shell.MethodTarget;
+import org.springframework.shell.standard.AbstractShellComponent;
+import org.springframework.shell.standard.ShellComponent;
+import org.springframework.shell.standard.ShellMethod;
+import org.springframework.shell.standard.ShellOption;
+
+@ShellComponent
+public class RegisterCommands extends AbstractShellComponent {
```
onobc, line 25: This does test the dynamic commands, but is not a test. 

I was looking for an IT that does some coarse-grained testing that we could add coverage to. We do not have one - may be a good thing to add. 

Do the current unit tests cover the new functionality?

PR #390, opened 2022-03-25
Refactoring to make the code readable and cohesive.

aravindjayanthi98:


onobc:
@aravindjayanthi98 thanks for the efforts but we typically do not accept wide reaching code proposals around formatting or re-arrangement as such. Targeting a focused improvement/refactoring/formatting when fixing an issue or adding a new feature is fine. 

Again, we do appreciate your efforts. There are many [issues](https://github.com/spring-projects/spring-shell/issues) available that we could use help on. 

Thanks

PR #396, opened 2022-05-02
Rework command subsystem

### LINKED ISSUE #380: Rework command subsystem
One trouble in #379 when I started to implement it is that while it's easy to make `CommandRegistry` writable and register new methods based on `MethodTarget` this process from a user perspective is a bit cumbersome.

Essentially we still need to make custom pojo:
```
	public static class PojoMethods {

		@ShellMethod
		public String dynamic1() {
			return "dynamic1";
		}

		@ShellMethod
		public String dynamic2(String arg1) {
			return "dynamic2" + arg1;
		}
	}
```
and have those methods with dummy `@ShellMethod` for `StandardParameterResolver` to read parameters.

jvalkeal:
- Focus of these changes are to introduce a new command system based on
  real registrations (new way) instead of continuously (old way) resolve
  methods and its parameters via reflection.
- There's a lot of changes as this resolution via reflection had its
  hooks almost everywhere and thus most changes are just refactorings.
- Order to understand real changes I'd start to look classes under
  `org.springframework.shell.command` package as it defines new registration,
  catalog and parser classes. Also samples contain new classes to demonstrate
  new functionality.
- Fixes #380


spring-shell-autoconfigure/src/main/java/org/springframework/shell/boot/CommandCatalogAutoConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+		methodTargetRegistrars.orderedStream().forEach(resolver -> {
+			resolver.register(catalog);
+		});
+		commandCatalogCustomizers.orderedStream().forEach(customizer -> {
+			customizer.customize(catalog);
+		});
+		return catalog;
+	}
+
+	@Bean
```
onobc, line 50: 1. I would consider ordering this default customizer as line44 spins through them ordered. 
2. Would we ever want to disable or override this default behavior?

spring-shell-core/src/main/java/org/springframework/shell/command/CommandCatalog.java
```diff
@@ -0,0 +1,189 @@
...
+
+	/**
+	 * Unregister a {@link CommandRegistration} by its command name.
+	 *
+	 * @param commandName the command name
+	 */
+	void unregister(String... commandName);
+
+	/**
+	 * Gets all {@link CommandRegistration}s mapped with their names.
```
onobc, line 60: I know this is a high level review - but I would eventually elaborate a bit more here as the default impl resolves dynamic commands as well as filter based on interactive/non-interactive. 

Also, I think its important to mention the map is a copy and any updates made to it will be ignored.

spring-shell-core/src/main/java/org/springframework/shell/command/CommandCatalog.java
```diff
@@ -0,0 +1,189 @@
...
+	Map<String, CommandRegistration> getRegistrations();
+
+	/**
+	 * Interface to resolve currently existing commands. It is useful to have fully
+	 * dynamic set of commands which may exists only if some conditions in a running
+	 * shell are met. For example if shell is targeting arbitrary server environment
+	 * some commands may or may not exist depending on a runtime state.
+	 */
+	@FunctionalInterface
+	interface CommandResolver {
```
onobc, line 73: Any reason to put these interfaces in here nested? I have found that nesting the interfaces in designs makes it harder to resonate with. 

I would strongly prefer the default impl to be outside of the interface definition as well.

spring-shell-core/src/main/java/org/springframework/shell/command/CommandCatalog.java
```diff
@@ -0,0 +1,189 @@
...
+	static CommandCatalog of(Collection<CommandResolver> resolvers, ShellContext shellContext) {
+		return new DefaultCommandCatalog(resolvers, shellContext);
+	}
+
+	/**
+	 * Default implementation of a {@link CommandCatalog}.
+	 */
+	static class DefaultCommandCatalog implements CommandCatalog {
+
+		private final HashMap<String, CommandRegistration> commandRegistrations = new HashMap<>();
```
onobc, line 122: Any particular reason to type to `HashMap` rather than `Map`?

spring-shell-core/src/main/java/org/springframework/shell/command/CommandCatalog.java
```diff
@@ -0,0 +1,189 @@
...
+		private final Collection<CommandResolver> resolvers = new ArrayList<>();
```
onobc, line 123: I am curious on the choice of Collection rather than Set or List here.

spring-shell-core/src/main/java/org/springframework/shell/command/CommandCatalog.java
```diff
@@ -0,0 +1,189 @@
...
+					}
+					else if (mim == InteractionMode.NONINTERACTIVE) {
+						return cim == InteractionMode.NONINTERACTIVE || cim == InteractionMode.ALL;
+					}
+					return true;
+				})
+				.collect(Collectors.toMap(Entry::getKey, Entry::getValue));
+		}
+
+		private static String commandName(String[] commands) {
```
onobc, line 181: I think it would be an improvement to  encapsulate this into a method in CommandRegistration. The Builder talks about this detail and it would be great if that behavior was w/ the CR itself. 

```java
		 * Define commands this registration uses. Essentially defines a full set of
		 * main and sub commands. It doesn't matter if full command is defined in one
		 * string or multiple strings as "words" are splitted and trimmed with
		 * whitespaces. You will get result of {@code command subcommand1 subcommand2, ...}.
```

spring-shell-core/src/main/java/org/springframework/shell/command/CommandCatalog.java
```diff
@@ -0,0 +1,189 @@
...
+		@Override
+		public Map<String, CommandRegistration> getRegistrations() {
+			HashMap<String, CommandRegistration> regs = new HashMap<>();
+			regs.putAll(commandRegistrations);
+			for (CommandResolver resolver : resolvers) {
+				regs.putAll(resolver.resolve());
+			}
+			return regs.entrySet().stream()
+				.filter(e -> {
+					InteractionMode mim = e.getValue().getInteractionMode();
```
onobc, line 165: At 1st glance, this hurts my head. I get it now, but it took too long. I have found in "big" filters like this creating a well-named private method and filtering by method reference makes it more readable.

spring-shell-autoconfigure/src/main/java/org/springframework/shell/boot/LineReaderAutoConfiguration.java
```diff
@@ -45,15 +45,15 @@ public class LineReaderAutoConfiguration {
 
 	private Parser parser;
 
-	private CommandRegistry commandRegistry;
+	private CommandCatalog commandRegistry;
```
onobc, line 48: I'll mention this once  here as its a **[NIT]** but worth pointing out... the var name needs to change as well. Its still reads "Registry" below in the code.

spring-shell-autoconfigure/src/main/java/org/springframework/shell/boot/ParameterResolverAutoConfiguration.java
```diff
@@ -1,24 +1,35 @@
...
 
 @Configuration(proxyBeanMethods = false)
 public class ParameterResolverAutoConfiguration {
 
 	@Bean
-	public ParameterResolver standardParameterResolver(ConversionService conversionService,
-			ObjectProvider<ValueProvider> valueProviders) {
-		Set<ValueProvider> collect = valueProviders.orderedStream().collect(Collectors.toSet());
-		return new StandardParameterResolver(conversionService, collect);
+	public CompletionResolver defaultCompletionResolver() {
```
onobc, line 22: Any need to disable or override this default?

spring-shell-standard/src/main/java/org/springframework/shell/standard/StandardMethodTargetRegistrar.java
```diff
@@ -74,11 +79,75 @@ public void register(ConfigurableCommandRegistry registry) {
...
-					commands.put(key, target);
+
+					Builder builder = CommandRegistration.builder()
+						.command(key)
+						.group(group)
+						.help(shellMapping.value())
+						.interactionMode(shellMapping.interactionMode())
+						.availability(availabilityIndicator);
+
+					InvocableHandlerMethod ihm = new InvocableHandlerMethod(bean, method);
```
onobc, line 92: Can you leave a comment here describing what all this is doing? I got lost on this one. :)

onobc (commented):
@jvalkeal I have taken a 1st pass but am going to read the docs and then go back through `spring-shell-core` changes in a  2nd pass. I will submit my current review comments though and continue more tomorrow.

spring-shell-core/src/main/java/org/springframework/shell/command/CommandCatalog.java
jvalkeal: Indeed, I've had same thoughts.

spring-shell-autoconfigure/src/main/java/org/springframework/shell/boot/CommandCatalogAutoConfiguration.java
jvalkeal: 2. Now that you mentioned it, we probably should have a flag.

spring-shell-core/src/main/java/org/springframework/shell/command/CommandCatalog.java
jvalkeal: This is a bit difficult design part of this interface. As you already commented an actual command name is and its use is spanning `CommandCatalog` and `CommandRegistration`. I originally had this just returning a list of registrations but then you need to calculate names out from registrations which felt weird.

spring-shell-core/src/main/java/org/springframework/shell/command/CommandCatalog.java
jvalkeal: ack

spring-shell-autoconfigure/src/main/java/org/springframework/shell/boot/LineReaderAutoConfiguration.java
jvalkeal: ack

spring-shell-autoconfigure/src/main/java/org/springframework/shell/boot/ParameterResolverAutoConfiguration.java
jvalkeal: Currently I don't think so.

spring-shell-core/src/main/java/org/springframework/shell/command/CommandCatalog.java
jvalkeal: Yes, and I think I'll polish interfaces around this.

jvalkeal:
I'm going to merge this. Think I've addresses most of a comments and reworked this pr. I'll create follow-up stories for some things which needs a bit more thinking.

jvalkeal:
Merged per 8a23518b84edc68c4a0e71ab4af47675fe410cbc

PR #468, opened 2022-07-18
Add the possibility of default selection to MultiItemSelector

Drevsh:
I hope I did everything correctly. (It's my first time committing to an open source project)

I think there might be a problem regarding default selection in the SingleItemSelector. 
You can now set multiple items as selected in the  SingleItemSelector and the DefaultSingleItemSelectorContext will return one at random.

```
		@Override
		public Optional<I> getResultItem() {
			if (getResultItems() == null) {
				return Optional.empty();
			}
			return getResultItems().stream().findFirst();
		}
```

Or that is at least what I think will happen. Maybe this needs to be handled differently in SingleItemSelector/MultiItemSelector

### DIFF
--- a/spring-shell-core/src/main/java/org/springframework/shell/component/flow/SelectItem.java
+++ b/spring-shell-core/src/main/java/org/springframework/shell/component/flow/SelectItem.java
@@ -43,11 +43,13 @@ public interface SelectItem {
 	 */
 	boolean enabled();
 
+	boolean selected();
+
 	public static SelectItem of(String name, String item) {
-		return of(name, item, true);
+		return of(name, item, true, false);
 	}
 
-	public static SelectItem of(String name, String item, boolean enabled) {
-		return new DefaultSelectItem(name, item, enabled);
+	public static SelectItem of(String name, String item, boolean enabled, boolean selected) {
+		return new DefaultSelectItem(name, item, enabled, selected);
 	}
 }
--- a/spring-shell-core/src/main/java/org/springframework/shell/component/flow/SelectItem.java
+++ b/spring-shell-core/src/main/java/org/springframework/shell/component/flow/SelectItem.java
@@ -43,6 +43,11 @@ public interface SelectItem {
 	 */
 	boolean enabled();
 
+	/**
+	 * Return if the item is selected.
+	 * 
+	 * @return if item is selected
+	 */
 	boolean selected();
 
 	public static SelectItem of(String name, String item) {
--- a/spring-shell-core/src/main/java/org/springframework/shell/component/support/Selectable.java
+++ b/spring-shell-core/src/main/java/org/springframework/shell/component/support/Selectable.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package org.springframework.shell.component.support;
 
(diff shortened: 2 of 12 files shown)


spring-shell-core/src/main/java/org/springframework/shell/component/support/Selectable.java
```diff
@@ -0,0 +1,6 @@
+package org.springframework.shell.component.support;
```
jvalkeal, line 1: Can you add license header.

spring-shell-samples/src/main/java/org/springframework/shell/samples/standard/ComponentCommands.java
```diff
@@ -93,7 +93,7 @@ public String singleSelector() {
 	public String multiSelector() {
 		List<SelectorItem<String>> items = new ArrayList<>();
 		items.add(SelectorItem.of("key1", "value1"));
-		items.add(SelectorItem.of("key2", "value2", false));
+		items.add(SelectorItem.of("key2", "value2", false, true));
```
jvalkeal, line 96: This would also need a change in https://github.com/spring-projects/spring-shell/blob/643b189fb8d63f9a7aeee361f12a30301864e9a2/spring-shell-core/src/main/resources/org/springframework/shell/component/multi-item-selector-default.stg#L21 as it was assumed that if item is disabled it would never show selected/disabled style. I'd also change below key3 to selected to demonstrate it with item which can be changed.

spring-shell-core/src/main/java/org/springframework/shell/component/flow/SelectItem.java
```diff
@@ -43,11 +43,13 @@ public interface SelectItem {
 	 */
 	boolean enabled();
 
+	boolean selected();
```
jvalkeal, line 46: Can you add javadocs.

jvalkeal (changes requested):
Thx, it looks something how I'd done the quick hack to add this feature without rewriting shared parent class with single/multi. It probably gives some bullets for a user to shoot own leg if this is used with single selector but I think we can address that later with additional changes or documentation.

I left some comments to address. Feel free to add yourself as author to files if you want. I think we can sneak this in into 2.1.0.

spring-shell-samples/src/main/java/org/springframework/shell/samples/standard/ComponentCommands.java
Drevsh: Do we need new styling here? So as far as I can see, if it is disabled we currently would only be using the "style-item-disabled", even it was selected right?

Do you have an idea how this could be styled? Currently there are the following for stages:
- enabled & selected
- enabled & not selected
- disabled 

instead of the disabled we would have:
- disabled & selected
- disabled & not selected

The disabled & not selected could be the same styling (as it currently is), but is is simply enough to add the checkbox if it is selected?

```
<if(item.enabled)>
  <if(item.selected)>
    <({<figures.checkboxOn> }); format=selected_style(item.selected)> <item.name>
  <else>
    <({<figures.checkboxOff> }); format=selected_style(item.selected)> <item.name>
  <endif>
<else>	
  <if(item.selected)>
     <({<figures.checkboxOn> }); format="style-item-disabled"> <item.name; format="style-item-disabled">
  <else>
    <({<figures.checkboxOff> }); format="style-item-disabled"> <item.name; format="style-item-disabled">
  <endif>  
<endif>
```


Could you also give me a hint how I can test the format changes? Building the application is quite cumbersome and takes some time just to test if the styling is appropriate.

jvalkeal: Yes it should be as simple as what you showed. Testing these are difficult as it's not easy when ansi escape characters are involved and `MultiItemSelectorTests` takes you only that far. I'd just try it out manually until we get something better for tests(I've been working on a node bases e2e tests which is sometimes only way to test what user is seeing).

Drevsh: image

Actually I think that looks quite nice.

Drevsh:
Hey @jvalkeal I hope I got everything.

jvalkeal (approved):

jvalkeal:
Thx a lot, merged per 2862cc855e05c7dcc5c6a9c42d632006a219eb7b

PR #494, opened 2022-08-01
Fix ShellMethod not using context validator

### LINKED ISSUE #493: Spring shell is defaulting to the base Jakarta validator
In the InvocableShellMethod, the doInvoke method reverts to the base Validator as opposed to the context one. This effectively prevents (Spring Boot from injecting beans into the JSR-303 Bean Validation API)[https://docs.spring.io/spring-framework/docs/3.0.0.RC1/reference/html/ch05s07.html].

After some debugging, it's been retraced to the following:
Image

This is most likely a leftover from debugging and should be omitted to allow for the class Validator to be used.

terminalsin:
Resolves #493


jvalkeal:
I think this is good for now, thx! We're probably going to rework these validations in 3.x but that's another story.

PR #530, opened 2022-09-14
Update: JLine Terminal should be configurable.

### LINKED ISSUE #516: JLine Terminal should be configurable
Currently `JLineShellAutoConfiguration` creates a default jline terminal which is fine for basic uses cases. It might be beneficial to make that bean conditional if user needs to change it and in #489 it's anyway required.

Koooooo-7:
close #516

### DIFF
--- a/spring-shell-autoconfigure/src/main/java/org/springframework/shell/boot/JLineShellAutoConfiguration.java
+++ b/spring-shell-autoconfigure/src/main/java/org/springframework/shell/boot/JLineShellAutoConfiguration.java
@@ -39,7 +39,6 @@ import org.springframework.shell.jline.PromptProvider;
  * @author Florent Biville
  */
 @Configuration(proxyBeanMethods = false)
-@ConditionalOnProperty(prefix = "spring.shell.jline-shell", value = "enabled", havingValue = "true", matchIfMissing = true)
 public class JLineShellAutoConfiguration {
 
 	@Bean(destroyMethod = "close")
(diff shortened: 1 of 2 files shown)


spring-shell-autoconfigure/src/main/java/org/springframework/shell/boot/JLineShellAutoConfiguration.java
```diff
@@ -38,9 +39,11 @@
  * @author Florent Biville
  */
 @Configuration(proxyBeanMethods = false)
+@ConditionalOnProperty(prefix = "spring.shell.jline-shell", value = "enabled", havingValue = "true", matchIfMissing = true)
```
jvalkeal, line 42: I think we don't need this setting?

Koooooo-7: > I think we don't need this setting?

Personally, I think if user wanna implement all the Beans under this configuration, we can provide a toggle to do it  instead of  loading this empty @Configuration/Component.

jvalkeal: Having flag in this level doesn't make sense as those beans are currently used elsewhere. Enabled flags are usually used in higher level feature sets.

Koooooo-7: > Having flag in this level doesn't make sense as those beans are currently used elsewhere. Enabled flags are usually used in higher level feature sets.

I see, gonna an update here.

Koooooo-7:
Hi @jvalkeal , any updates?

PR #532, opened 2022-09-14
feat: support component level prefix.

### LINKED ISSUE #531: [Enhancement]  Support component level prefix.
Normally, we collect the related command under the same `@ShellComponent`,  and we may wanna add a common prefix for each command instead `concat` it in each `@ShellMethod` key.
```java
	@ShellComponent(prefix = "docker-prefix")
	public static class ShellComponentWithPrefix {

		@ShellMethod(value = "run")
		public void foo1(...) {
		}
                 
                 @ShellMethod(value = "build")
		public void foo2(...) {
		}

	}
```
It should be 
```
docker-prefix run
docker-prefix build
```

Koooooo-7:
close #531


Koooooo-7:
@jvalkeal  PTAL.

PR #612, opened 2023-01-11
Allow Shell.input() to be overridden in subclasses

marvin-enthus:
Hi,

can we set `Shell.input()` to `protected`, so it can be overriden by other projects?

In https://github.com/fonimus/ssh-shell-spring-boot is a class [ExtendedShell.java](https://github.com/fonimus/ssh-shell-spring-boot/blob/main/starter/src/main/java/com/github/fonimus/ssh/shell/ExtendedShell.java#L110) which needs to override and also access the input()-Method to inject input received via the SSH session properly.

Thanks,
     Chris

### DIFF
--- a/spring-shell-core/src/main/java/org/springframework/shell/Shell.java
+++ b/spring-shell-core/src/main/java/org/springframework/shell/Shell.java
@@ -190,7 +190,7 @@ public class Shell {
 	 * result
 	 * </p>
 	 */
-	private Object evaluate(Input input) {
+	protected Object evaluate(Input input) {
 		if (noInput(input)) {
 			return NO_INPUT;
 		}


jvalkeal:
I think it should be fine.

We'be been cleaning up `Shell` class quite a bit as it used to contain way too much logic and probably will go away at some point. Might be good idea to start thinking what users might need trying to modify anything `Shell` class has to get better api's and extension points.

marvin-enthus:
Hi @jvalkeal Thanks' for your feedback. Just my cents regarding the api / extension: I'm also thinking of a 'web interface' to spring shell. It might have similar requirements like the spring-ssh extension: session management (as the web UI could be opened serveral times by different users), permission management and a different way to inject the input as it comes from the web console. Quick'n'dirty way would be using a web-based SSH terminal client 🤣

PR #693, opened 2023-03-19
Pass ShellContext to CommandCatalog.of in CommandCatalogAutoConfig

cachescrubber:
## Pass ShellContext to CommandCatalog.of in CommandCatalogAutoConfiguration

A Small Bugfix. The `ShellContext` is not passed to `CommandCatalog.of` in `CommandCatalogAutoConfiguration`. As a consequence, command filtering by interaction mode is not working as expected.


jvalkeal:
Ah good catch thanks. Looks like this broke in a refactoring commit 8a23518b84edc68c4a0e71ab4af47675fe410cbc.

Issue #512, opened 2022-08-20
CompletionProposal with complete


fonimus:
Hello, I am providing in my application an ExtendedFileValueProvider which does not stop if comes a directory (i.e. the completion proposal is not complete). Today I had to extended some classes to do this, and I am wondering if it could be included in spring shell.

The aim is to add `complete` boolean field in `CompletionProposal` in order to be able to get this value in `CompleterAutoConfiguration.CompleterAdapter#complete` :

````
proposals.stream()
				.map(p -> new Candidate(
					p.dontQuote() ? p.value() : cpl.emit(p.value()).toString(),
					p.displayText(),
					p.category(),
					p.description(),
					null,
					null,
					true) --> here we can set p.complete() instead of true
				)
				.forEach(candidates::add);
````

This way the class implementing `ValueProvider` can decide if the proposal is complete or not.
Do you think it could be an improvement in spring shell library ?

Thanks a lot !

jvalkeal:
Do you have screenshot or something similar to show how it looks like?

fonimus:
Sure !
In this gif example I have written the following command :

````
@ShellMethod("File command")
public void file(
            @ShellOption(valueProvider = FileValueProvider.class, defaultValue = ShellOption.NULL) File file,
            @ShellOption(valueProvider = ExtendedFileValueProvider.class, defaultValue = ShellOption.NULL) File extended
) {
        info(file);
        info(extended);
}
````

and we can see that when I set complete=false, the cursor stays put and I can press tab again to follow path ; if I use existing file provider I have to press del key to go back and then press tab.

![extended-file-provider](https://user-images.githubusercontent.com/26186612/186601563-e53237f0-b076-4edb-8c71-390e38f3afd6.gif)

jvalkeal:
That would be nice addition. If you have a time please make a PR and we can continue from there!

tincore:
@jvalkeal Just created this quick PR for this issue. https://github.com/spring-projects/spring-shell/pull/715

Not to diverge too much. I think that positional parameters do not autocomplete with tab. is this intended?

Issue #285, opened 2019-08-30
Tab auto-complete will always put a space after selected option


gaetPax:
Currently when selecting an option through auto-complete (implemented with ValueProvider) the selected option will always put a a space in the command line after the selected option.
This is not always the desired behaviour, especially when autocompleting a file path (i.e. using the FileValueProvider). You would like to go deeper in the path and stay on the last character.

I think think this occurs because the 'complete' field of the 'Candidate' Jline class is always set to true in the 'CompleterAdapter' class:
https://github.com/spring-projects/spring-shell/blob/master/spring-shell-core/src/main/java/org/springframework/shell/jline/JLineShellAutoConfiguration.java#L198

This behaviour should be parametrised.

tincore:
This always true assignment seems to currently be at `CompleterAutoConfiguration`

`
		@Override
		public void complete(LineReader reader, ParsedLine line, List<Candidate> candidates) {
			CompletingParsedLine cpl = (line instanceof CompletingParsedLine) ? ((CompletingParsedLine) line) : t -> t;

			CompletionContext context = new CompletionContext(sanitizeInput(line.words()), line.wordIndex(), line.wordCursor(), null, null);

			List<CompletionProposal> proposals = shell.complete(context);
			proposals.stream()
				.map(p -> new Candidate(
					p.dontQuote() ? p.value() : cpl.emit(p.value()).toString(),
					p.displayText(),
					p.category(),
					p.description(),
					null,
					null,
					true) // <-----  COMPLETE ALWAYS TRUE 
				)
				.forEach(candidates::add);
		}
`

PR #715, opened 2023-04-15
Option for auto-completing single argument without spaces insertion (complete support)

tincore:
Support for completing  single argument with multiple tab clicks. For example file paths.

I've also adjusted the File completion to add a separator when completion reaches directory. Then mark as complete when reaches a file.

I quickly looked and could not find related test cases to add this scenarios. 

Related issues:
https://github.com/spring-projects/spring-shell/issues/285
https://github.com/spring-projects/spring-shell/issues/512


jvalkeal:
Merged to main. Going to look backporting to maintenance branches.

PR #844, opened 2023-08-04
feat: adding mandatory flag to string input

Nico-DF:
As I began to use spring-shell for some CLI workflow, I found that there is no input required for string (altough there is default).

For the moment, I used this kind of code:
```java
public String input(String message, String defaultValue, boolean mandatory) {
	var component = new StringInput(getTerminal(), message, defaultValue);
	component.setResourceLoader(getResourceLoader());
	component.setTemplateExecutor(getTemplateExecutor());
	var context = component.run(ConfirmationInput.ConfirmationInputContext.empty());
	var result = context.getInput();

	while (mandatory && !StringUtils.hasText(result)) {
		result = component.run(ConfirmationInput.ConfirmationInputContext.empty()).getInput();
	}

	return result;
} 
```

or this as flow:
```java
componentFlowBuilder.clone().reset() //
    .withStringInput("id")
	.name("Select an identifier")
	.next(ctx -> StringUtils.isBlank(ctx.getInput()) ? "id" : null)
    .and().build();
```

Which work but it could be simpler, hence my PR.

So I:
- Added to flow the mandatory flag
- Added to component the mandatory flag
- Add simple test case
- Add samples

Result:
image
image
image


Nico-DF:
Issue with jdk8 compilation doesn't seem to originate from me, right?

Nico-DF:
As suspected I broke something :D, I'll check it

spring-shell-core/src/main/java/org/springframework/shell/component/StringInput.java
```diff
@@ -26,11 +26,9 @@
 import org.jline.utils.AttributedString;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
 import org.springframework.shell.component.StringInput.StringInputContext;
 import org.springframework.shell.component.context.ComponentContext;
 import org.springframework.shell.component.support.AbstractTextComponent;
-import org.springframework.shell.component.support.AbstractTextComponent.TextComponentContext;
```
jvalkeal, line 33: This is causing compile failure with jdk8

spring-shell-samples/src/main/java/org/springframework/shell/samples/standard/ComponentCommands.java
```diff
@@ -56,6 +56,15 @@ public String stringInput(boolean mask) {
 		return "Got value " + context.getResultValue();
 	}
 
+	@ShellMethod(key = "component string required", value = "String input", group = "Components")
```
jvalkeal, line 59: I'd probably combine this with existing `stringInput` by adding a new flag instead of creating a new command.

spring-shell-core/src/main/java/org/springframework/shell/component/StringInput.java
Nico-DF: Indeed, my IDE removed it as unused, reverting it

spring-shell-samples/src/main/java/org/springframework/shell/samples/standard/ComponentCommands.java
Nico-DF: also hesitated, will do

Nico-DF: But I will remove the default value, otherwise it doesn't make sense

jvalkeal:
I'm just thinking if this is going to be a bit confusing for a user if there's a default value.

```
my-shell:>component string required
? Enter value [Required][Default myvalue]
```

I'm kinda questioning if there's better way to give indication to user that something is required.

Nico-DF:
> I'm just thinking if this is going to be a bit confusing for a user if there's a default value.
> 
> ```
> my-shell:>component string required
> ? Enter value [Required][Default myvalue]
> ```
> 
> I'm kinda questioning if there's better way to give indication to user that something is required.

I also wondered about it. I think the easiest way would be to display Required only if there is no default (and it also matches with what really happens)

fmbenhassine:
@Nico-DF Can you please rebase this PR on the latest `main`? I will check if we can include it in the upcoming 4.0. Thank you upfront.

Nico-DF:
Update and signing done

fmbenhassine:
Thank you for the quick updates! You were faster than I expected 😂 I was in the middle of a big refactoring when I asked for a rebase and I didn't think your update would come the same day..

Anyway, I will try to rebase your PR myself, otherwise I will let you know.

Nico-DF:
No problem, was in between tasks at work, so I  had time to do it. I can rebase it tomorrow morning if you want, just ping me when all is finished on main

fmbenhassine:
Thank you very much! But no rush, we can merge this (as well as #849) in RC1 next week. I will drop a message here when `main` is stable again (some restructuring is ongoing).

fmbenhassine:
Hi @Nico-DF 

I think we can proceed and merge this PR now. Can you please rebase it on the latest `main`?

Please exclude the changes in samples, they are outdated.

Thank you upfront.

Nico-DF:
Done 👍 
Not had time to compile locally yet 😋

fmbenhassine:
Thank you for the updates! 

> Not had time to compile locally yet 😋

I tired locally and it compiles! Ship it 😂 2bdf8432af0038e6759761c2ed86a75471dd1a6e

PR #875, opened 2023-10-02
Update StringTemplate version in gradle.properties

eedijs:
Newer version contains bugfixes.


eedijs:
This version also contains license in `pom.xml`:
```xml
<licenses>
    <license>
        <name>The BSD License</name>
        <url>http://www.antlr.org/license.html</url>
        <distribution>repo</distribution>
    </license>
</licenses>
```

Issue #943, opened 2023-12-02
Disable 


stellirin:

stellirin:
After further testing it seems that, at least in some cases, both properties are required to fully disable a feature. History is easy to test when run as a user without permission to write to `spring-shell.log`:

**Test 1**

No properties set. `history` command is enabled and app does seem to use `spring-shell.log`

- `help` command works but errors with `java.nio.file.AccessDeniedException: /**/spring-shell.log`
- `history` command visible in `help`
- `history` command is enabled, errors with `java.nio.file.AccessDeniedException: /**/spring-shell.log`

**Test 2**

```
spring.shell.history.enabled=false
```

`history` command is enabled but app doesn't seem to use `spring-shell.log`

- `help` command works without issue
- `history` command visible in `help`
- `history` command is enabled, returns `[]`

**Test 3**

```
spring.shell.command.history.enabled=false
```
`history` command is not enabled but app does seem to use `spring-shell.log`

- `help` command works but errors with `java.nio.file.AccessDeniedException: /**/spring-shell.log`
- `history` command **not** visible in `help`
- `history` command is not enabled, returns `No command found for 'history'`

**Test 4**

```
spring.shell.history.enabled=false
spring.shell.command.history.enabled=false
```

`history` command is not enabled and app doesn't seem to use `spring-shell.log`

- `help` command works without issue
- `history` command **not** visible in `help`
- `history` command is not enabled, returns `No command found for 'history'`

jvalkeal:
Looks like we should document these better. If looking _history_, `spring.shell.history.enabled` and `spring.shell.command.history.enabled` are two different things. Former disabling history functionality in an interactive(i.e arrows up/down) shell and latter disabling the history command itself. You don't need to have `history command` to have _history_ feature enabled in an interactive mode.

PR #1065, opened 2024-05-04
Correcting application property name in history command docs

jelovac:


jelovac:
The documentation is misleading. It appears that both of these properties are required to disable history as stated in https://github.com/spring-projects/spring-shell/issues/943

Closing this PR as I was not aware of this.

Issue #277, opened 2019-04-03
Quotes Should Not Be Recognized In Comments


jlevy-bigfootbiomedical:
Both double and single quotes are recognized (or "active") inside of comments, which I think is a serious bug.   For example:

The following two line Spring Shell script will not print help:
// don't do this
help

And if you put those two lines in a file, and try to run it with the at-sign on the command line, it will crash and you will get this: 
Caused by: org.jline.reader.EOFError: Missing closing quote

dougbot01:
I also have found this to be a problem, especially when running script files as mentioned.

Is there a way to up-vote a fix for this? If someone can point to where this might be broken (spring shell vs jLine) maybe I can dig into a fix.

polo7:
Dear colleagues, I've created a [PR to fix this issue](https://github.com/spring-projects/spring-shell/pull/1087). Could you please approve merging if there are no objections.

PR #1087, opened 2024-06-08
Parse commented lines like an empty lines

polo7:
Fixes [GH-277](https://github.com/spring-projects/spring-shell/issues/277).
Idea: we parse a line and then evaluate despite it is commented. If there is a quote inside of comment then we get an exception "Missing a closing quote" (line 141 in spring-shell-core/src/main/java/org/springframework/shell/jline/ExtendedDefaultParser.java). To avoid this behavoir we should treat commented line like an empty. Then evaluate() from spring-shell-core/src/main/java/org/springframework/shell/Shell.java will take care of it with its "fail fast" on line 195.


polo7:
Hello, @jvalkeal ! I didn't want to disturb you but it seems you are the only active official in this repo. Could you please guide me with my further steps? I guess I need your approval here.

polo7:
> Thank you for the PR! LGTM 👍
> 
> Can you please rebase it on the latest `main` branch and add a test for that case? It should then be good to merge for the upcoming 3.4.1 release.
> 
> Thank you upfront.

@fmbenhassine I've rebased my branch on top of `main` and added unit tests to cover different cases and different implementations of Parser. Let me know if there is anything else I need to do to go further.

fmbenhassine:
Thank you for the update! However I don't see the PR rebased on the latest `main` with those 48 commits and 109 files changed (not related to this PR). The PR should have a single commit with the change set you want to introduce. I don't know how I can help here but try to rebase your PR again. Otherwise you can open another one if it's easier for you.

polo7:
@fmbenhassine I went through PR creation once again and it shows me PR is already opened with changes I've made.
Maybe I should close this one and create PR one more time, I don't know. It seems like a bug here.
Снимок экрана 2025-04-29 в 11 24 05

polo7:
@fmbenhassine this is my another PR instead of this one.

Issue #979, opened 2024-01-18
Support arg=value format with options


Martiul:
Hello,

I recently upgraded from spring-shell 2.7.5 to 3.2.0, and found that arguments of the form "arg=value" are no longer accepted.

Is this a regression, or was it intended?

Sample code:
```
  @ShellMethod(
      value = "Description",
      key = {"test"})
  public void test(
      @ShellOption(
          value = {"--foo", "-f"},
          defaultValue = ShellOption.NULL) String foo) {
    System.out.println("Foo is: " + foo);
  }
```

`shell:> test --foo=bar`

Expected:
```
Foo is: bar
```

Actual:

```
Unrecognised option '--foo=bar'
```

jvalkeal:
Thanks for pointing this out. It wasn't implemented when parser infra went through complete overhaul.

I'm just thinking that when we add support for this should we be explicit which formats are supported, `--arg value` vs. `--arg=value`. It might be convenient to allow user to choose either or both.

Martiul:
Thanks for the response @jvalkeal .  For me specially, this is not a blocking/urgent bug, but figured it was worth mentioning in case other users do the same migration

fmbenhassine:
This is fixed in Spring Shell 4 which now follows the same parsing rules as Spring Boot (options can be specified with key=value pairs), with enhanced capabilities following the POSIX style, see: https://docs.spring.io/spring-shell/reference/4.0-SNAPSHOT/commands/syntax.html

Martiul:
thanks!

PR #1144, opened 2024-10-05
Start of key value implementation

emileplas:
This is related to https://github.com/spring-projects/spring-shell/issues/979

I spent some time this afternoon with a very simple start: 

- added method to separate key=value pairs and to add key and value to the argument list.
- added some basic initial tests
- added configuration option to enable or disable the key value separation

I, however, have some additional questions:

**1. For array values: currently, that would look like:**
```
command option one two three
```
With a key-value separator: would that look like this?
 ```
command option==one two three
```
**2. empty values:**
The current (initial start) implementation would assign this as an empty value:
```
command option=
```
I can see where that could be unclear. 

**3. place of current key-value implementation:**
Currently, the key-value split method will look _at_ all arguments, including commands. The advantage is that it doesn't touch the Lexer.class tokenize method, but I would understand if _only_ options should be separated from their value using the key value separator. In that case, I will change this a bit.

I didn't want to push this initial _draft_ too far, before getting your feedback. Please let me know what you think.

### DIFF
--- a/spring-shell-core/src/main/java/org/springframework/shell/command/parser/Lexer.java
+++ b/spring-shell-core/src/main/java/org/springframework/shell/command/parser/Lexer.java
@@ -77,6 +77,58 @@ public interface Lexer {
 		private record ArgumentsSplit(List<String> before, List<String> after) {
 		}
 
+
+		/**
+		 * Loops over a list of arguments and if any of the arguments would follow the pattern ".*=.*"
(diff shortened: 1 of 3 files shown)


fmbenhassine:
As mentioned in #979, specifying options with key=value pairs is now supported in v4. This PR is outdated with the recent changes in v4.

Thank you for your time anyway!

PR #1168, opened 2025-04-29
Parse commented lines like an empty lines

polo7:
This is re-PR due to some bugs (original: https://github.com/spring-projects/spring-shell/pull/1087)

Idea: we parse a line and then evaluate despite it is commented. If there is a quote inside of comment then we get an exception "Missing a closing quote" (line 141 in spring-shell-core/src/main/java/org/springframework/shell/jline/ExtendedDefaultParser.java). To avoid this behavoir we should treat commented line like an empty. Then evaluate() from spring-shell-core/src/main/java/org/springframework/shell/Shell.java will take care of it with its "fail fast" on line 195.

### DIFF
--- a/spring-shell-core/src/main/java/org/springframework/shell/jline/FileInputProvider.java
+++ b/spring-shell-core/src/main/java/org/springframework/shell/jline/FileInputProvider.java
@@ -66,7 +66,13 @@ public class FileInputProvider implements InputProvider, Closeable {
         if (line == null) {
             return null;
         } else {
-            ParsedLine parsedLine = parser.parse(sb.toString(), sb.toString().length());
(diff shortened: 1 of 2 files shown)


polo7:
@fmbenhassine Hey there! This my new PR instead of https://github.com/spring-projects/spring-shell/pull/1087
Is it fine now? I hope it is :)

fmbenhassine:
LGTM, thank you for the PR!

I will amend a small change, see my previous comment.

fmbenhassine:
Rebased and merged as 25f02d54f57a42c1e3b69b48d227568b4eb4dc51. Refined in 30531c8cbb2c466698ce083180ba2932cfb92eb4.

Issue #1167, opened 2025-04-24
Improve command aliasing


fmbenhassine:
As of v3.4, creating an [alias](https://docs.spring.io/spring-shell/reference/commands/alias.html) for a command is done as follows:

```java
CommandRegistration commandRegistration() {
	return CommandRegistration.builder()
		.command("mycommand")
		// define alias as myalias
		.withAlias()
			.command("myalias")
			.and()
		// define alias as myalias1 and myalias2
		.withAlias()
			.command("myalias1", "myalias2")
			.and()
		.build();
}
```

The DSL methods `command("alias").and()` are repetitive. I suggest to shorten them as follows:

```java
CommandRegistration commandRegistration() {
	return CommandRegistration.builder()
		.command("mycommand")
		// define a single alias
		.withAlias("myalias")
		// or define multiple ones at once
		.withAlias("myalias1", "myalias2")
		.build();
}
```

fmbenhassine:
I didn't see that it was also possible to alias the group:

```java
[..]
.withAlias()
   .command("alias1")
   .group("Alias Group")
   .and()
[..]
```

So the shortened syntax I suggested previously does not make sense anymore (how to pass the group as part of the new method signature `.withAlias("myalias1", "myalias2")`?)

I think the best way to improve command aliasing would be to resolve it as part of #1127 (which BTW mentions aliasing: `[..] This could be done for withOption, withTarget, withAlias, withErrorHandling, withHelpOption and withExitCode [..]`), something like:

```java
// before
CommandRegistration commandRegistration() {
	return CommandRegistration.builder()
		.command("mycommand")
		// define alias
		.withAlias()
			.command("myalias")
                        .group("myalias group")
			.and()
		.build();
}

// after
CommandRegistration commandRegistration() {
	return CommandRegistration.builder()
		.command("mycommand")
		// define alias
		.withAlias(aliasSpec - > aliasSpec.command("myalias").group("myalias group"))
		.build();
}
```

PR #1170, opened 2025-05-07
Improve command aliasing in `CommandRegistration.BaseBuilder`

therepanic:
I think we should also add a method in the builder with varargs for `withAlias` to make things easier. Here I also added a test related to `withAlias` where I also added a test for our new `withAlias(String... commands)` method.


fmbenhassine:
Thank you for the PR! However, I think the change in this PR  won't work since it's possible to specify the alias group as well, see https://github.com/spring-projects/spring-shell/issues/1167#issuecomment-3467529144.

Hence I am closing this PR. Thank you anyway for your time! 🙏

PR #1188, opened 2025-10-03
Add spring code formatting plugin to the build

therepanic:
In this change, we add support for the `spring-javaformat-maven-plugin` plugin. Please note that in addition to adding the plugin itself, I ran the formatting process and only that. I believe that to verify this, you can click on the format on your local version and check that all changes match this commit.

Closes: gh-1185


piotrooo:
@fmbenhassine I think we should apply the code formatting ASAP. This helps with further refactors. WDYT?

fmbenhassine:
Thank you for the PR! Rebased and merged as 328f23a546ff1959768dfe43578109e0bb43be0a.

I just amended a version update of the plugin and removed the `disable.format.checks` flag (not sure we will ever disable that flag in practice as the goal is to always make sure the code is correctly formatted).

Issue #1203, opened 2025-10-31
Build successful despite errors


fmbenhassine:
Several tests pass in the IDE and on the CLI despite errors, for example: `MethodCommandExceptionResolverTests` succeeds even if there is the following error:

```
[main] WARN org.springframework.shell.core.command.annotation.MethodCommandExceptionResolver - Failure in @ExceptionResolver
java.lang.NullPointerException: Cannot invoke "Object.getClass()" because "providedArgs[providedArgsIndex]" is null
	at org.springframework.shell.core.command.invocation.InvocableShellMethod.getMethodArgumentValues(InvocableShellMethod.java:273)
	at org.springframework.shell.core.command.invocation.InvocableShellMethod.invoke(InvocableShellMethod.java:231)
	at org.springframework.shell.core.command.annotation.MethodCommandExceptionResolver.resolve(MethodCommandExceptionResolver.java:88)
	at org.springframework.shell.core.command.annotation.MethodCommandExceptionResolverTests.shouldErrorWhenResolving(MethodCommandExceptionResolverTests.java:117)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:701)
	at org.junit.platform.commons.support.ReflectionSupport.invokeMethod(ReflectionSupport.java:502)
	at org.junit.jupiter.engine.support.MethodReflectionUtils.invoke(MethodReflectionUtils.java:45)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:61)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$ValidatingInvocation.proceed(InvocationInterceptorChain.java:124)
	at org.junit.jupiter.engine.extension.TimeoutExtension.intercept(TimeoutExtension.java:163)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestableMethod(TimeoutExtension.java:148)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestMethod(TimeoutExtension.java:86)
	at org.junit.jupiter.engine.execution.InterceptingExecutableInvoker$ReflectiveInterceptorCall.lambda$ofVoidMethod$0(InterceptingExecutableInvoker.java:123)
[...]

PR #1208, opened 2025-11-02
Fix running shell in the Spring Boot applications

piotrooo:
Partially resolves: gh-1203

After the removal of `DefaultShellApplicationRunner` in this commit: https://github.com/spring-projects/spring-shell/commit/ff73bdccb9dfb6f9c6645c1bf0d71e8fa2d40c8e, Spring Boot applications no longer have an entry point for shell runners.

This causes shell programs not to run.

### DIFF
--- a/spring-shell-autoconfigure/src/main/java/org/springframework/shell/boot/ApplicationRunnerAutoConfiguration.java
+++ b/spring-shell-autoconfigure/src/main/java/org/springframework/shell/boot/ApplicationRunnerAutoConfiguration.java
@@ -15,20 +15,36 @@
  */
 package org.springframework.shell.boot;
 
+import org.springframework.beans.factory.ObjectProvider;
+import org.springframework.boot.ApplicationArguments;
+import org.springframework.boot.ApplicationRunner;
 import org.springframework.boot.autoconfigure.AutoConfiguration;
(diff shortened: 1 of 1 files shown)


fmbenhassine:
@piotrooo Thank you for the PR! But I think there is deeper issue: command discovery is still broken. In the hello world example, `spring-shell-starter` brings `spring-shell-standard` which is supposed to add basic commands, but that is not the case. I mean with this PR, the hello world sample now works again and the command `hello` is recognized, but it's impossible to exit the shell.. `exit` and `quit` are not registered.

Moreover, I do not understand why `ShellRunner#run` returns a boolean and not `void` as it was before. The javadoc mentions `Return value indicates if run operation happened and no further runners should be used.` but in my opinion it is not the ShellRunner's repsonsbility to decide the flow of execution (ie if another runner should run or not). More explicitly, If I want to implement a custom runner, I don't know if I should return true or false in my custom implementation. In others words, having to do the following seems awkward to me:

```java
boolean run = runner.run(args.getSourceArgs());
if (run) {
   return;
}
```

I think we need to fix that before proceeding as well.. Thoughts?

piotrooo:
I've opened a discussion at https://github.com/spring-projects/spring-shell/discussions/1209. It should be more convenient.

Issue #1186, opened 2025-09-22
Make interactive mode the default


fmbenhassine:
#342 added non-interactive mode, but also made it the default.

Most typical Spring Shell apps are interactive by design, and users should opt-in the non-interactive mode rather than opt-out.

PR #1210, opened 2025-11-05
Make interactive mode the default

therepanic:
In the current commit, I'm changing the runner creation rules in the autoconfiguration module. Now, if the user doesn't specify a specific runner, we create a bean with `InteractiveShellRunner`.

I also rewrote the configuration test to accommodate our change.

Closes: gh-1186

**Note**: the build breaks not because of the current change, but because the previous commit in the repository breaks the build.


fmbenhassine:
This was introduced as part of 32587727eee6bee5b6a560049c02acbd4424c320 and d5eb4c355f4abe92a0fb341f15cf17e619b436e5. Those were quite large change sets and I couldn't rebase your PR (due to outdated changes, apologies).

Thank you for your your time and effort anyway 🙏

Issue #1215, opened 2025-11-07
Revisit dependency to apache commons-io


fmbenhassine:
Currently, `spring-shell-core` has a compile time dependency to `commons-io` for a single usage in `org.springframework.shell.core.tui.component.PathSearch`. From what I see, this usage can be replaced with the recent Java IO APIs added in the JDK. This issue is to update `PathSearch` to use the Java APIs and remove the dependency to `commons-io`.

Contributions are welcome!

therepanic:
Hi, @fmbenhassine! Can you assign it to me? I want work on it

fmbenhassine:
@therepanic Sure! thank you for your offer to help 🙏

PR #1222, opened 2025-11-17
Remove commons-io dependency from PathSearch

therepanic:
In this commit, we replace the use of commons-io in PathSearch with api from jdk.

Closes: gh-1215


fmbenhassine:
Amazing! Thank you very much for this PR 👍 

Rebased and merged as f791234d040c2ef1b2cec1ac48606017486fa2dd. I just removed [this part](https://github.com/spring-projects/spring-shell/blob/c4f86aa6fd3d833d7f321e35d0f261f05faaea7b/spring-shell-core/src/test/java/org/springframework/shell/core/tui/component/PathSearchTests.java#L53-L61) from the test as it is outdated.

Issue #1213, opened 2025-11-07
Revisit dependency to reactor


fmbenhassine:
I could not find detailed context about this decision, but I see no added value of using reactor in the context of shell applications (both in interactive and non interactive modes, with all the complexity currently in `DefaultEventLoop`). Might be related to #1205 as well.

I would argue to remove this dependency for a more streamlined core module.

Issue #1205, opened 2025-11-02
The reason of Spring Messaging  in Spring Shell


fmbenhassine:
### Discussed in https://github.com/spring-projects/spring-shell/discussions/1204

<div type='discussions-op-text'>

<sup>Originally posted by **piotrooo** November  1, 2025</sup>
I'd like to ask about the rationale for using `spring-messaging` in `spring-shell`. The only reason I can see is to reuse `Message` and `HandlerMethodArgumentResolver` as some kind of generic holder (?) for multiple purposes.

Is that really the best approach? It feels  a bit bizarre.

Maybe a `spring-shell` specific `HandlerMethodArgumentResolver` should be introduced, and the dependency on `spring-messaging` could be dropped?</div>

fmbenhassine:
I couldn't agree more! I came to the same conclusion when I was reviewing dependencies (similar to having `core` depending on a `ui` module, see c95651abca97d756bdbafc8d0a6105a90e1928ac).. Same thing for reactor dependency, I see no added value of using it in the context of interactive shell applications (even in non interactive mode BTW, with all the complexity in `DefaultEventLoop`).

Anyway, I will plan this for v4 and see if we can improve things with less dependencies.

PR #1230, opened 2025-12-23
Move Spring Messaging and Reactor dependencies from core to jline

piotrooo:
As this dependency is used only in the `jline` module, it should reside there.

This is the very first step to mitigate the usage of Spring Messaging and Reactor (#1213 and #1205).

Additionally, an `.editorconfig` file was added to help with indentation for users who use spaces instead of tabs.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,5 @@
+root = true
+
+[*.{java,xml,gradle}]
+indent_style = tab
+indent_size = 4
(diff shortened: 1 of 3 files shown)


fmbenhassine:
This is great! I like how the core module is streamlined with less dependencies. I did something similar for ST4 as well: 70a37c433ddb120e5a256a330e62ea4124589cb6

Reabsed and merged as 1b30182b9625c31e01e2bea2f262728774bc4618. Thank you for the PR 🙏

PR #1232, opened 2025-12-23
Remove Spring Messaging dependency

### LINKED ISSUE #1205: The reason of Spring Messaging  in Spring Shell

piotrooo:
Resolve #1205 

This is the first step toward removing Spring Messaging and Reactor dependencies.

The `Message` class from the Spring Messaging project was used only to leverage message header functionality, which was used in a dynamic manner.

The `Message` object has been replaced with an internal `TerminalEvent` object that mimics all required header features.

> [!IMPORTANT]
> This change is a pure object replacement intended to remove the Spring Messaging dependency. No logic or control flow has been changed. Further refactoring will be done in future iterations.


fmbenhassine:
This is absolutely amazing! Thank you for doing this 🙏 The `TerminalEvent` concept and name are great! I believe this should go hand in hand with the reactor replacement.

I will review this in details and see if we can include it in v4.0 planned for next week (might be short at this point). My thinking is that if there are no reactor APIs "leaked" in Spring Shell public APIs, then we can leave the reactor replacement for 4.1 or later.

piotrooo:
> This is absolutely amazing! Thank you for doing this 🙏 The `TerminalEvent` concept and name are great! I believe this should go hand in hand with the reactor replacement.
> 
> 
> 
> I will review this in details and see if we can include it in v4.0 planned for next week (might be short at this point). My thinking is that if there are no reactor APIs "leaked" in Spring Shell public APIs, then we can leave the reactor replacement for 4.1 or later.

As I mentioned, baby steps — let’s start with Spring Messaging and go deeper after the 4.0 release 🤔 Reactor is unfortunately not trivial, but maybe you have some ideas 💡

fmbenhassine:
> As I mentioned, baby steps — let’s start with Spring Messaging and go deeper after the 4.0 release

That's what I am saying, the idea is to see if we can merge this now for 4.0 and keep the reactor replacement for 4.1 if no breaking changes could be expected. 

However, reactor APIs are leaked all over the places in many Spring Shell public APIs.. This is really unfortunate.. The good news is that reactor usage seems to be only in the Terminal UI framework, which is documented as experimental and subject to breaking changes: https://docs.spring.io/spring-shell/reference/3.4/tui/index.html . So we can safely work on removing these two dependencies (at the same time as I believe they go hand in hand) in a minor release like 4.1 or 4.2.

Issue #1246, opened 2025-12-26
Cannot define completion only for a particular command option


czpilar:
Consider this configuration:

```java
@SpringBootApplication
public class SpringShellApplication {

    public static void main(String[] args) {
        SpringApplication.run(SpringShellApplication.class, args);
    }

    @Command(name = "hello", completionProvider = "helloNameCompletionProvider")
    public void sayHello(@Option(longName = "name", shortName = 'n') String name) {
        System.out.println("Hello " + name + "!");
    }

    @Bean
    public CompletionProvider helloNameCompletionProvider() {
        return completionContext -> {
            CommandOption option = completionContext.getCommandOption();
            if (option == null) {
                return Collections.emptyList();
            }
            if ("name".equals(option.longName()) || 'n' == option.shortName()) {
                return Stream.of("Peter", "Paul", "Mary").map(CompletionProposal::new).toList();
            }
            return Collections.emptyList();
        };
    }

}
```

Step to reproduce:
1. Type `hello`
2. Press `Tab` and select `--name` or `-n` option
3. Press `Tab` again and observe that no completion proposal is shown.

This is caused because no command option is provided within completion context so `completionContext.getCommandOption()` is always `null`.

czpilar:
Resolved in PR by providing `CommandOption` to `CompletionContext`.

czpilar:
Solving also situation when one long option name is defined as `--name` and another as `--na`.

PR #1250, opened 2025-12-28
Improved completion by allowing configuration of whether option values should be completed

### LINKED ISSUE #285: Tab auto-complete will always put a space after selected option

czpilar:
Improves #1246
Resolves #285

### DIFF
--- a/spring-shell-core/src/main/java/org/springframework/shell/core/command/CommandOption.java
+++ b/spring-shell-core/src/main/java/org/springframework/shell/core/command/CommandOption.java
@@ -28,7 +28,8 @@ import org.springframework.util.StringUtils;
  * @author David Pilar
  */
 public record CommandOption(char shortName, @Nullable String longName, @Nullable String description,
-		@Nullable Boolean required, @Nullable String defaultValue, @Nullable String value, Class<?> type) {
+		@Nullable Boolean required, @Nullable String defaultValue, @Nullable String value, Class<?> type,
+		boolean completion) {
 
 	public boolean isOptionEqual(String optionName) {
 		return StringUtils.hasLength(longName) && optionName.equals("--" + longName)
@@ -54,6 +55,8 @@ public record CommandOption(char shortName, @Nullable String longName, @Nullable
 		private @Nullable String value;
 
 		private Class<?> type = Object.class;
+
+		private boolean completion = true;
 
 		public Builder shortName(char shortName) {
 			this.shortName = shortName;
@@ -90,8 +93,13 @@ public record CommandOption(char shortName, @Nullable String longName, @Nullable
 			return this;
 		}
 
+		public Builder completion(boolean completion) {
+			this.completion = completion;
+			return this;
+		}
+
 		public CommandOption build() {
-			return new CommandOption(shortName, longName, description, required, defaultValue, value, type);
+			return new CommandOption(shortName, longName, description, required, defaultValue, value, type, completion);
 		}
 
 	}
--- a/spring-shell-core/src/main/java/org/springframework/shell/core/command/annotation/Option.java
+++ b/spring-shell-core/src/main/java/org/springframework/shell/core/command/annotation/Option.java
@@ -62,4 +62,10 @@ public @interface Option {
 	 */
 	String defaultValue() default "";
 
+	/**
+	 * Indicates whether this option should be completed in the command completer.
+	 * @return true if the option should be completed, defaults to true.
+	 */
+	boolean completion() default true;
+
 }
--- a/spring-shell-core/src/main/java/org/springframework/shell/core/command/annotation/support/CommandFactoryBean.java
+++ b/spring-shell-core/src/main/java/org/springframework/shell/core/command/annotation/support/CommandFactoryBean.java
@@ -135,6 +135,7 @@ public class CommandFactoryBean implements ApplicationContextAware, FactoryBean<
 				String description = optionAnnotation.description();
 				boolean required = optionAnnotation.required();
 				String defaultValue = optionAnnotation.defaultValue();
+				boolean completion = optionAnnotation.completion();
 				CommandOption commandOption = CommandOption.with()
 					.shortName(shortName)
 					.longName(longName)
@@ -142,6 +143,7 @@ public class CommandFactoryBean implements ApplicationContextAware, FactoryBean<
 					.required(required)
 					.defaultValue(defaultValue)
 					.type(parameter.getType())
+					.completion(completion)
 					.build();
 				commandOptions.add(commandOption);
 			}
--- a/spring-shell-jline/src/main/java/org/springframework/shell/jline/CommandCompleter.java
+++ b/spring-shell-jline/src/main/java/org/springframework/shell/jline/CommandCompleter.java
@@ -47,11 +47,14 @@ public class CommandCompleter implements Completer {
 				// add option completions for the command
 				for (CommandOption option : options) {
 					boolean present = isOptionPresent(line, option);
+					String separator = option.completion() ? "" : "=";
 					if (StringUtils.hasLength(option.longName()) && !present) {
-						candidates.add(new Candidate("--" + option.longName()));
+						candidates.add(new Candidate("--" + option.longName() + separator, "--" + option.longName(),
+								null, null, null, null, option.completion(), 0));
 					}
 					if (option.shortName() != ' ' && !present) {
-						candidates.add(new Candidate("-" + option.shortName()));
+						candidates.add(new Candidate("-" + option.shortName() + separator, "-" + option.shortName(),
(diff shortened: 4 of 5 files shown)


czpilar:
Hello,

I’ve just improved `--key=value` completion by parameterizing the `Option` annotation so you can specify whether the option value should be completed or not. By default, this is set to `true`, which means option names are completed with a space. When you specify `completion=false`, an `=` is appended to the option name instead. This improves #1246.

```java
@Option(longName = "first", shortName = 'f', completion = false) String first
```

I’ve also improved the completion provider so you can control whether a completion proposal should be completed or not, as mentioned in #285.

See the following examples.
The first example shows two commands where the `hello` command has options with completion enabled. The second command, `hello2`, has options with completion disabled.

This means completion will behave as follows:
```
shell:> hello --first Peter --last Chan

shell:> hello2 --first=Peter --last=Chan
```

```java

import org.springframework.shell.core.command.CommandOption;
import org.springframework.shell.core.command.completion.CompletionProposal;
import org.springframework.shell.core.command.completion.CompletionProvider;

import java.util.Collections;
import java.util.stream.Stream;

@SpringBootApplication

	}

	@Command(name = "hello", completionProvider = "helloNameCompletionProvider")
	public void sayHello(@Option(longName = "first", shortName = 'f') String first,
			@Option(longName = "last", shortName = 'l') String last) {
		System.out.println("Hello " + first + " " + last + "!");
	}

	@Command(name = "hello2", completionProvider = "helloNameCompletionProvider")
	public void sayHello2(@Option(longName = "first", shortName = 'f', completion = false) String first,
			@Option(longName = "last", shortName = 'l', completion = false) String last) {
		System.out.println("Hello2 " + first + " " + last + "!");
	}

	@Bean
	public CompletionProvider helloNameCompletionProvider() {
		return completionContext -> {
			CommandOption option = completionContext.getCommandOption();
			if (option == null) {
			}

			String word = completionContext.getWords().get(completionContext.getWords().size() - 1);
			String prefix = word.contains("=") ? word.substring(0, word.indexOf('=') + 1) : "";

			Stream<String> options = Stream.empty();

			if ("first".equals(option.longName())) {
				options = Stream.of("Peter", "Paul", "Mary");
			}
			else if ("last".equals(option.longName())) {
				options = Stream.of("Chan", "Noris");
			}

			return options.map(str -> new CompletionProposal(prefix + str).displayText(str)).toList();
		};
	}

}
```

The second example shows how to use a completion provider with completion disabled for its completion proposals.

This means completion will behave as follows:
```
shell:>snake --first=Mary/Paul/Paul/Paul/Paul/Paul/Paul/Paul/Paul/Paul --last=Noris/Chan/Chan/Noris/Noris/Noris
```

In this case, the completion proposals are appended with `/`, and the proposals themselves are not completed. This was requested in #285.

```java

import org.springframework.shell.core.command.CommandOption;
import org.springframework.shell.core.command.completion.CompletionProposal;
import org.springframework.shell.core.command.completion.CompletionProvider;

import java.util.Collections;
import java.util.stream.Stream;

@SpringBootApplication

	}

	@Command(name = "snake", completionProvider = "snakeCompletionProvider")
	public void snake(@Option(longName = "first", shortName = 'f', completion = false) String first,
			@Option(longName = "last", shortName = 'l', completion = false) String last) {
		System.out.println("Snake " + first + " " + last + "!");
	}

	@Bean
	public CompletionProvider snakeCompletionProvider() {
		return completionContext -> {
			CommandOption option = completionContext.getCommandOption();
			if (option == null) {
			}

			String word = completionContext.getWords().get(completionContext.getWords().size() - 1);
			String prefix = word.endsWith("=") || word.endsWith("/") ? word : word + "/";

			Stream<String> options = Stream.empty();

			if ("first".equals(option.longName())) {
				options = Stream.of("Peter", "Paul", "Mary");
			}
			else if ("last".equals(option.longName())) {
				options = Stream.of("Chan", "Noris");
			}

			return options.map(str -> new CompletionProposal(prefix + str).displayText(str).complete(false)).toList();
		};
	}

}
```

I’ve also added many tests to cover this behavior.

As a next step, it could be useful to allow defining a custom separator character instead of just a space or equals sign, but that would require changes to the command parser as well.

czpilar:
@fmbenhassine Please take a look at this PR.

czpilar:
Converting this to a draft PR. There will be conflicts with #1256. After #1256 is merged, I will resolve the conflicts.

czpilar:
Conflicts resolved, it is ready to merge if the change is suitable. Please, let me know.

fmbenhassine:
Thank you for the PR!

I need more time to think about this new `completion` attribute you suggest here. Please keep this open and I will take a look after 4.0.1.

czpilar:
@fmbenhassine Yes, I’m aware of the new attribute. I’m absolutely fine if you suggest a different solution or attribute name, or if you decide this isn’t a good approach at all. I’ll keep this PR open and resolve all conflicts.

fmbenhassine:
> I’ve just improved `--key=value` completion by parameterizing the `Option` annotation so you can specify whether the option value should be completed or not.

We typically would like to see all available options and be able to complete them when hitting <kbd>Tab</kbd>. Now, since we already have an extension point that makes this possible (ie a custom `CompletionProvider` that ignores options we want to ignore), there is no point in adding a feature to do the same thing through the annotation's attribute. Do you see my point? I always try to keep things as simple as possible and have a single way to do things.

> Resolves https://github.com/spring-projects/spring-shell/issues/285
> I’ve also improved the completion provider so you can control whether a completion proposal should be completed or not, as mentioned in #285.

I think #285 is not valid with v4 as the completion flag is not hardcoded to `true` anymore now but instead it is [taken](https://github.com/spring-projects/spring-shell/blob/af799eda675f281dfc8751f0bf085c597f859898/spring-shell-jline/src/main/java/org/springframework/shell/jline/CommandCompleter.java#L65) from the proposal, which the user creates/controls in the implementation of `CompletionProvider`. So I think that issue could be solved with a custom provider as well, no need for an additional feature in the framework. Here is a quick example I tried (not even with a custom completion provider, but with the built-in file provider to mimic the use case of the OP):

Replace the Boot sample with this:

```java

import java.io.IOException;
import java.nio.file.Files;
import java.nio.file.Path;
import java.nio.file.Paths;
import java.util.stream.Stream;

import org.springframework.shell.core.command.completion.CompletionProvider;
import org.springframework.shell.core.command.completion.FileNameCompletionProvider;

@SpringBootApplication

	}

	@Command(name = "cat", completionProvider = "catCompletionProvider")
	public void cat(@Option(longName = "file", shortName = 'f', required = true) String file) {
		Path path = Paths.get(file);
		try (Stream<String> lines = Files.lines(path)) {
			lines.forEach(System.out::println);
		}
		catch (IOException e) {
			throw new RuntimeException(e);
		}
	}

	@Bean
	public CompletionProvider catCompletionProvider() {
		return new FileNameCompletionProvider();
	}

}
```

When completing the "file" option, I can press <kbd>Tab</kbd> several times to go deeper in the path and Spring Shell does not add a space when I select a folder, here is an example:

```
shell:>cat --file .github/workflows/ci.yml
name: CI

on:
  workflow_dispatch:
  push:
    branches:
      - main
    paths-ignore:
      - '.github/**'

jobs:
  build:
    runs-on: ubuntu-latest

[...] the rest of the file
```

What do you think? Did I miss a detail?

> As a next step, it could be useful to allow defining a custom separator character instead of just a space or equals sign, but that would require changes to the command parser as well.

I am not sure we need to go that deep in the configurability of the parser. Key/value pairs separated by ' ' or '=' are the most common ways of specifying options. If the user wants to deviate from that, she/he can always provide a custom parser (which is now possible with v4, but was impossible with v3).

czpilar:
Hi @fmbenhassine,

Thank you for the thorough review and the detailed explanation!

I submitted this proposal about 3 months ago, and a lot has changed in the codebase since then - so some parts of it may already be obsolete.

Looking at it now with fresh eyes, I'm also sceptical about introducing a new annotation attribute for such a non-standard use case, especially given that the existing `CompletionProvider` extension point already covers this scenario. Your point about keeping things simple and having a single way to do things makes sense to me.

I'm positive about closing this PR as **not planned / won't do**. Feel free to close it.

Thanks again for the time you put into reviewing this!

PR #1256, opened 2025-12-30
Fix not working completion in some cases and using the display text of a completion proposal

### LINKED ISSUE #1255: Completion does not work correctly in some cases, and display text is not used in candidates
**4.0.0 GA**

Hello,

I've found some cases where completion of options, option values and commands does not work correctly, and display text is not used in candidates.

Consider a configuration where `.displayText(str)` is defined on a `CompletionProposal`.

```java

import org.springframework.shell.core.command.CommandOption;
import org.springframework.shell.core.command.completion.CompletionProposal;
import org.springframework.shell.core.command.completion.CompletionProvider;

import java.util.Collections;
import java.util.stream.Stream;

@SpringBootApplication

    }

    @Command(name = "hello", completionProvider = "helloCompletionProvider")
    public void hello(@Option(longName = "first", shortName = 'f') String first,
                      @Option(longName = "last", shortName = 'l') String last) {
        System.out.println("Hello " + first + " " + last + "!");
    }

    @Bean
    public CompletionProvider helloCompletionProvider() {
        return completionContext -> {
            CommandOption option = completionContext.getCommandOption();
            if (option == null) {
            }

            String word = completionContext.getWords().get(completionContext.getWords().size() - 1);
            String prefix = word.contains("=") ? word.substring(0, word.indexOf('=') + 1) : "";

            Stream<String> options = Stream.empty();

            if ("first".equals(option.longName())) {
                options = Stream.of("Peter", "Paul", "Mary");
            }
            else if ("last".equals(option.longName())) {
                options = Stream.of("Chan", "Noris");
            }

            return options.map(str -> new CompletionProposal(prefix + str).displayText(str)).toList();
        };
    }

}
```
## Completion does not work correctly in some cases

1. Type `hello --first Pe` and press `Tab` ---> it does not complete the option value
    ```
    shell:>hello --first Pe<TAB>
    ```
2. Type `hello --first` and press `Tab` ---> it does not complete the option
    ```
    shell:>hello --first<TAB>
    ```
3. Type `hello` and press `Tab` ---> it does not complete the command
    ```
    shell:>hello<TAB>
    ```

## Display text is not used in candidates

Also, when using a command completion provider, the display text of a `CompletionProposal` is not used to display the completion candidate. Instead, the actual value passed to the `CompletionProposal` constructor is used.

1. Type `hello --first=` and press `Tab`.
2. The completion proposals are shown as `--key=value` pairs instead of just the display text (e.g., `Peter`, `Paul`, `Mary`).

Image

The issue is that I would like the completion candidates to be displayed the same way as they are without the equals sign, as shown in the following screenshot:

Image

czpilar:
Resolves #1255

### DIFF
--- a/spring-shell-jline/src/main/java/org/springframework/shell/jline/CommandCompleter.java
+++ b/spring-shell-jline/src/main/java/org/springframework/shell/jline/CommandCompleter.java
@@ -15,6 +15,7 @@ import org.springframework.shell.core.command.completion.CompletionProvider;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.function.Predicate;
 
 /**
  * A JLine {@link Completer} that completes command names from a {@link CommandRegistry}.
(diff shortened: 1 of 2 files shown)


fmbenhassine:
LGTM 👍 Rebased, squashed and merged as 17f79eb0f61e8bc51e34146b90a87609684b1763.

PR #1283, opened 2026-01-06
Fix command highlighter

### LINKED ISSUE #1277: Alias commands are not coloured correctly after update to version 4.0.0
I build my shell application with the jansi starter and have defined a command as follows:
```
   @Command(description = "Gets state of running AMELIO job", name = "get-jobstate", alias = "gjs", group = "Job Commands")
   public String getState() {
```

When I type "get-jobstate" into the shell the command is shown in white:

Image

When I use the alias name "gjs" ist will be displayed in red:

Image

The command is accepted and executed. It is only a problem with the colouring.

piotrooo:
Resolves #1277 

Bonus:
* Cleanup `pom.xml` (tabs)
* Add missing `package-info.java`
* Align tests to JUnit 6


fmbenhassine:
Rebased and merged as 03c1f47ac34e6b7ca8496e9ef4de4e18d418e7bd. Refined in 5272ecb19fce617d94a2a9c020eeb7c4424dfb4e (minor changes mentioned above).

I did not merge bonus points as I want to keep commits minimal and focused on single issues, but those are good to take in another change set.

PR #1293, opened 2026-01-24
Fix quoting of option values is broken #1291

### LINKED ISSUE #1291: Quoting of option values is broken
In version 3.x you could execute a command using a quoted option value like that: 

`select --from some_table --where "counter=7"`

In version 4.0.0, the quotes are ignored and the value for the where option is parsed to `counter`. 

I did some debugging and found an issue in the class `org.springframework.shell.core.command.DefaultCommandParser`. In the `parseOption` method, the string containing the option name and the option value is split using `word.split("=")`. This splits the quoted value, too. This seems unnecessary, as the entire input string was already split using a regular expression which checks for quoted parts of the string. Using `word.split("=", 2)` should return the correct result.

Additionally, I would recommend to make the regular expression used in the `parse` method a bit more robust by changing it from `" (?=(?:[^"]*"[^"]*")*[^"]*$)"` to `"\s+(?=(?:[^"]*"[^"]*")*[^"]*$)"`. This would match one or more spaces.

czpilar:
Resolves #1291

### DIFF
--- a/spring-shell-core/src/main/java/org/springframework/shell/core/command/DefaultCommandParser.java
+++ b/spring-shell-core/src/main/java/org/springframework/shell/core/command/DefaultCommandParser.java
@@ -52,7 +52,7 @@ public class DefaultCommandParser implements CommandParser {
 	@Override
 	public ParsedInput parse(String input) {
 		log.debug("Parsing input: " + input);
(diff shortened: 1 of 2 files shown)


fmbenhassine:
LGTM 👍 Rebased and merged as 90ea5743ef2e08520034ee43d21e4834839c7570.

PR #1294, opened 2026-01-24
Improve error reporting when non mandatory options are misconfigured

### LINKED ISSUE #1245: Improve error reporting when non mandatory options are misconfigured or incorrectly specified at runtime
This seems like regression from some of the latests commits.

Consider this configuration:

```java
@SpringBootApplication

    }

    @Command(name = "hello", completionProvider = "helloNameCompletionProvider")
    public void sayHello(@Option(longName = "name", shortName = 'n') String name) {
        System.out.println("Hello " + name + "!");
    }

    @Bean
    public CompletionProvider helloNameCompletionProvider() {
        return completionContext -> Stream.of("Peter", "Paul", "Mary").map(CompletionProposal::new).toList();
    }

}
```

There are some issues:

1. If a command is invoked without any option, the warning is logged. Which is not consistent if command is invoked with and option without value (there is no warning message).
    ```
    shell:>hello
    11:32:19.249 [main] WARN  o.s.s.c.c.a.MethodInvokerCommandAdapter - No value provided for optional option '--name' and no default value specified.
    Hello !
   
    shell:>hello --name
    Hello !

    shell:>hello -n
    Hello !
    ```

2. If option is specified with a short or long name with space before the value, it's not recognized.
    ```
    shell:>hello --name Paul
    Hello !
   
    shell:>hello -n Paul
    Hello !
   ```
   
3. The value is recognized only if specified with equal sign.
    ```
    shell:>hello --name=Paul
    Hello Paul!

    shell:>hello -n=Paul
    Hello Paul!
    ```

Image

jayychoi:
This PR improves error reporting for optional options that are misconfigured.

- **Fail fast for optional options without default value**: `CommandFactoryBean` now validates that optional options must have a default value. If an optional option is declared without a default value, the application fails at startup with a clear error message instead of logging a warning at runtime.

- **Option without value already handled**: This is already handled by `DefaultCommandParser` since commit 78d1a47, which throws an `IllegalArgumentException` when parsing such input.

## Changes

- `CommandFactoryBean`: Added validation to reject optional options without default values at command registration time
- `CommandFactoryBeanTests`: Added unit tests for the new validation logic

Resolves #1245

### DIFF
--- a/spring-shell-core/src/main/java/org/springframework/shell/core/command/annotation/support/CommandFactoryBean.java
+++ b/spring-shell-core/src/main/java/org/springframework/shell/core/command/annotation/support/CommandFactoryBean.java
@@ -53,6 +53,7 @@ import org.springframework.util.Assert;
  * @author Janne Valkealahti
  * @author Piotr Olaszewski
  * @author Mahmoud Ben Hassine
+ * @author Jay Choi
  */
 public class CommandFactoryBean implements ApplicationContextAware, FactoryBean<Command> {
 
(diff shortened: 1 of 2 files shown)


fmbenhassine:
This PR was valid based on my comment here: https://github.com/spring-projects/spring-shell/issues/1245#issuecomment-3693003845.

However, as explained in https://github.com/spring-projects/spring-shell/issues/1245#issuecomment-4206451012, things have changed since then and we now accept non-mandatory options without default values. Therefore, this PR is not up-to-date anymore. I am closing it.

Thank you for your contribution anyway 🙏

PR #1295, opened 2026-01-24
Fix HelpTests to work also on windows

czpilar:
There is a common issue with line separators between Windows and Linux systems. I’ve fixed the tests to ignore line separators.


czpilar:
@fmbenhassine Please look at this PR, it would make easier to build spring shell on windows machine...

PR #1297, opened 2026-01-24
Introduced default completion provider supporting enumeration values if option is of enum type

### LINKED ISSUE #1282: Provide completion from enumeration values if an option is of enum type
It seems that in v3, when an option is of type enum, the default completer provided values from the enum's values. This makes sense and it seems to be a regression in v4.

This issue is to align the default completer of v4 with v3.

czpilar:
Resolves #1282


fmbenhassine:
This is fantastic! The idea of leveraging the `EnumCompletionProvider` in a default completion provider is great. Thank you for another great contribution 👍

PR #1302, opened 2026-01-28
Fixed script command to accept the script file as an option #1301

### LINKED ISSUE #1301: Revert script command to accept the script file as an option rather than an argument
The script command should be reverted to accept the script file as an option `--file` (as in v3) instead of an argument (as in v4). It's way more "natural" to type `script --file /home/user/my.script` than `script -- /home/user/my.script`.

czpilar:
Resolves #1301

### DIFF
--- a/spring-shell-core/src/main/java/org/springframework/shell/core/command/Script.java
+++ b/spring-shell-core/src/main/java/org/springframework/shell/core/command/Script.java
@@ -17,6 +17,7 @@ package org.springframework.shell.core.command;
 
 import java.io.File;
 import java.util.List;
+import java.util.Objects;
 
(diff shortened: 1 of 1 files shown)


fmbenhassine:
Thank you for the PR! LGTM 👍 

However, we need to override `Command#getOptions` to specify the new file option (needed for the help string), and also update the docs [here](https://docs.spring.io/spring-shell/reference/commands/builtin/script.html). But no problem, I will take care of those on merge.

fmbenhassine:
Rebased and merged as d4c1135cccd4b71803c920078af4a6189d72bcf7. I amended the two minor changes I mentioned above.

PR #1303, opened 2026-01-29
Fix longName defaulting to parameter name when not provided #1261

### LINKED ISSUE #1261: Commands fail in 4.0 if no short name or long name are specified
When upgrading to Spring Shell 4.0, all annotated command options that don't have a short name or long name specified fail the application initialisation.

If that change was intentional, it would be great to highlight it in the [migration guide](https://github.com/spring-projects/spring-shell/wiki/v4-migration-guide) as a breaking change and update the JavaDocs as the [`@Option`](https://github.com/spring-projects/spring-shell/blob/3.4.x/spring-shell-core/src/main/java/org/springframework/shell/command/annotation/Option.java#L34) annotation states that those two arguments are optional.

Still, it would be great to restore the previous behaviour where the Java property name was used as the long name by default. That would be in line with other Spring projects like Spring MVC, where the `@RequestParam` doesn't require specifying a name if it matches the Java property name.

czpilar:
Resolves #1261

### DIFF
--- a/spring-shell-core/src/main/java/org/springframework/shell/core/command/adapter/MethodInvokerCommandAdapter.java
+++ b/spring-shell-core/src/main/java/org/springframework/shell/core/command/adapter/MethodInvokerCommandAdapter.java
@@ -136,16 +136,11 @@ public class MethodInvokerCommandAdapter extends AbstractCommand {
 				log.debug("Processing option for parameter: " + parameters[i].getName());
 				char shortName = optionAnnotation.shortName();
 				String longName = optionAnnotation.longName();
-				if (shortName == ' ' && longName.isEmpty()) {
-					throw new IllegalArgumentException(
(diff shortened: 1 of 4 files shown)


fmbenhassine:
I was working on this as well and had almost the same change set!

[gh1261.patch](https://github.com/user-attachments/files/24954356/gh1261.patch)

LGTM 👍 Rebased and merged as 6b1e7016ab5af5a0ecb3cc24bdaee3bfac6d2d5d.

czpilar:
Yes, that is unfortunate - both of us working on the same problem. I wasn’t sure whether you had already started on this, but since the 4.0.1 release is planned for today, I took a look at it.

Moreover, I fixed the option proposal in `CommandCompleter` when the `longName` of an option is empty, as it was proposing `--`. After fixing the default value of `longName`, however, the empty value should no longer occur.

Issue #1267, opened 2026-01-02
Configuration properties metadata are missing in 4.0


ThomasVitale:
It seems like the metadata for the Spring Boot configuration properties defines in Spring Shell 4.0 are not generated correctly, and therefore not included in the published artefacts. That means there's no validation or autocompletion support from IDEs.

Example:

<img width="673" height="223" alt="Image" src="https://github.com/user-attachments/assets/051be861-486d-447a-8a69-c0cd4c0c78e5" />

fmbenhassine:
Thank you for reporting this.

> I'm available to help with a potential PR.

Yes please. I am not familiar with how meta-data are published, so if you know how to do that, I would be grateful if you contribute a PR! Many thanks upfront.

ThomasVitale:
I'll look into it and prepare a PR soon.

ThomasVitale:
@fmbenhassine I have raised a PR with a fix: https://github.com/spring-projects/spring-shell/pull/1306 Don't know if it's too late for including that in 4.0.1.

fmbenhassine:
Thank you very much! No it's not too late, I will include it in 4.0.1.

PR #1306, opened 2026-01-30
fix: Configuration properties metadata are missing

ThomasVitale:
The Maven compiler is now configured to allow running the Spring Boot Configuration Processor. Furthermore, a dependency to Spring Boot Autoconfigure Processor has been added to the autoconfigure module so that metadata is also generated for the autoconfiguration conditions.

Fixes gh-1267

References from the Spring Boot documentation: [here](https://docs.spring.io/spring-boot/specification/configuration-metadata/annotation-processor.html#page-title) and [here](https://docs.spring.io/spring-boot/reference/features/developing-auto-configuration.html#features.developing-auto-configuration.custom-starter.autoconfigure-module).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -141,6 +141,16 @@
 							<artifactId>nullaway</artifactId>
 							<version>${nullaway.version}</version>
 						</path>
+						<path>
+							<groupId>org.springframework.boot</groupId>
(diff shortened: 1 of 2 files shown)


fmbenhassine:
That's great! Rebased and merged as fe786cf717992d2e2ea5478d540f9febdbc2db9b.

PR #1307, opened 2026-01-31
Fixed help option in command

czpilar:


fmbenhassine:
LGTM 👍 We can move that check a little upper to avoid asking the registry with the help options. There is another [small optimization](https://github.com/spring-projects/spring-shell/commit/84abfda3#diff-61713bb33ce507221932562065fed82be2b602a671ea7fef88232f71a479d957L59-L64) that I seem to have removed inadvertently as well. I will amend those on merge.

Thank you very much, much appreciated 🙏

fmbenhassine:
Merged as fe806a011712ad70d08ad700569049f12e795c9a.

Great that you added a test as well 👍

PR #1310, opened 2026-02-01
Fix Boolean option without value does not properly work with argument #1309

### LINKED ISSUE #1309: Boolean option without value does not properly work with argument
As it was mentioned in the comment here boolean option without value does not properly work with argument:
https://github.com/spring-projects/spring-shell/issues/1304#issuecomment-3831679837

Considering this configuration:

```java

import org.springframework.shell.core.command.annotation.Argument;

@SpringBootApplication

	}

	@Command(name = "rm")
	public String rm(@Option(shortName = 'f', defaultValue = "false") boolean force, @Argument(index = 0) String name) {
		return "Removed " + name + " force " + force;
	}

}
```

It does not work as `shell:>rm --force filename`
But it works as `shell:>rm filename --force`

Image

czpilar:
Resolves #1309

### DIFF
--- a/spring-shell-core-autoconfigure/src/main/java/org/springframework/shell/core/autoconfigure/ShellRunnerAutoConfiguration.java
+++ b/spring-shell-core-autoconfigure/src/main/java/org/springframework/shell/core/autoconfigure/ShellRunnerAutoConfiguration.java
@@ -21,6 +21,7 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.context.annotation.Bean;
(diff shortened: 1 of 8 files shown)


fmbenhassine:
Thank you for opening the issue and contributing a fix!

The change set LGTM 👍 I just find the introduction of the conversion service a bit disruptive/intrusive (see its impact on breaking some public APIs) for what it is used for at the moment (just converting a boolean value). Since I am targeting this fix for a patch release,  I will update that conversion logic to a simpler version for now, and we can introduce the conversion service in a later minor iteration if needed.

fmbenhassine:
Rebased and merged as 0ece35ecacffdc4cc9492fec8389d94bfee80a8a.

As mentioned previously, I kept only the fix in the parser without the introduction of a conversion service. I adapted the `isBooleanValue` function to a simpler version accordingly.

czpilar:
@fmbenhassine Sure, simplifying is not a problem. I used the ConversionService because it is used in other parts of the project, and I thought it was the preferred way to convert values.

Issue #1314, opened 2026-02-08
Default option int value is ignored (in 4.0.1)


georgmittendorfer:
## Description

Specifying a default value for an int option does not work.

## Current Behavior

`@Option(longName = "...", description = "...", defaultValue = "25") @Positive int someOption`

Running a command without specifying the option results in

```
The following constraints were not met:
	--someOption: must be greater than 0
```

(Removing the `@Positive` and manually checking the value confirms that it is `0`.)

## Wanted behavior

Default value 25 should be set if the option is not specified.

## System Information

Linux
Version: Spring Shell 4.0.1
Interactive shell mode

This did work with previous versions, at least with 3.x.

georgmittendorfer:
I found where the problem is:

MethodInvokerCommandAdapter:
```
Class<?> parameterType = parameterTypes[i];
// check if the option type is primitive or not
if (!parameterType.isPrimitive()) {
	// try to convert default value if present
	String defaultValue = optionAnnotation.defaultValue();
	if (!defaultValue.isEmpty()) {
		Object value = this.conversionService.convert(defaultValue, parameterType);
		args.add(value);
	}
	else {
		// for non-primitive types, add null
		args.add(null);
	}
}
else {
	// for primitive types, add default value of the primitive
	args.add(Utils.getDefaultValueForPrimitiveType(parameterType));
}
```

Utils:
```
public static Object getDefaultValueForPrimitiveType(Class<?> type) {
	Assert.isTrue(type.isPrimitive(), "Type must be a primitive type");
	if (type == boolean.class) {
		return false;
	}
	else if (type == char.class) {
		return '\u0000';
	}
	else if (type == byte.class || type == short.class || type == int.class || type == long.class) {
		return 0;
	} // otherwise it's float or double
	return 0.0;
}
```

If a non primitive type like `Integer` instead of `int` is used it works as expected.

PR #1317, opened 2026-02-08
Fix defaultValue handling for primitive Option parameters

therepanic:
Fixes a case where defaultValue was ignored for primitive option types.

Closes: gh-1314

### DIFF
--- a/spring-shell-core/src/main/java/org/springframework/shell/core/command/adapter/MethodInvokerCommandAdapter.java
+++ b/spring-shell-core/src/main/java/org/springframework/shell/core/command/adapter/MethodInvokerCommandAdapter.java
@@ -45,6 +45,7 @@ import org.springframework.util.MethodInvoker;
  *
  * @author Mahmoud Ben Hassine
  * @author David Pilar
+ * @author Andrey Litvitski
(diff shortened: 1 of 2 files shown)


spring-shell-core/src/main/java/org/springframework/shell/core/command/adapter/MethodInvokerCommandAdapter.java
```diff
@@ -157,23 +158,21 @@ private List<Object> prepareArguments(CommandContext commandContext) {
...
 							}
 							else {
 								// for non-primitive types, add null
 								args.add(null);
 							}
 						}
-						else {
-							// for primitive types, add default value of the primitive
-							args.add(Utils.getDefaultValueForPrimitiveType(parameterType));
-						}
```
therepanic: Currently, we don't use a default value for primitives. I've changed that here, so now we use a default value for both primitives and non-primes and then check whether it's empty or not.

fmbenhassine:
LGTM 👍 rebased and merged as db567c26bab7dab5fbccc2b57d2fa3106108ba57.

Thank you for  your contribution!

PR #1319, opened 2026-02-10
Added JLine NonInteractiveShellRunner configuration #1312

### LINKED ISSUE #1312: NonInteractiveShellRunner v3 -> v4 Migration Issue
I am encountering an issue with a Spring Shell v4 non-interactive application.  When running the application in a Windows terminal, output that contains UTF-8 characters outside the BMP displays those characters as question marks.

I believe that this is because the NonInteractiveShellRunner uses System.console for output which uses the terminal's default character set (IBM437).  The JLine library supports writing unicode characters out to the console, so using it resolves this issue.

This is a Spring Shell v3 to v4 Migration issue.  Under Spring Shell v3, non-interactive applications still used JLine for console output but under Spring Shell v4, System.console is used.

It would be nice if the JLine library made a non-interactive ShellRunner class available to replace the default NonInteractiveShellRunner in order to make the JLine capabilities available when the JLine starter is used.

The following code demonstrates the issue in Spring Shell v4:
```java
package io.github.rmcdouga;

@SpringBootApplication
public class DemoApplication {

		SpringApplication.run(DemoApplication.class, args);
	}

	@Command(name = "example")
	public String example() {
		return "Hello World! \u2665";	// Should display a heart symbol but displays question mark instead.
	}
}
``` 

The equivilent code works under Spring Shell v3:
```java
package io.github.rmcdouga;

import org.springframework.shell.command.annotation.Command;
import org.springframework.shell.command.annotation.CommandScan;

@SpringBootApplication
@CommandScan
@Command
public class DemoApplication {

		SpringApplication.run(DemoApplication.class, args);
	}

	@Command(command = "example")
	public String example() {
		return "Hello World! \u2665"; // Displays a heart symbol
	}
}
```

czpilar:
Resolves #1312

### DIFF
--- a/spring-shell-core-autoconfigure/src/main/java/org/springframework/shell/core/autoconfigure/ShellRunnerAutoConfiguration.java
+++ b/spring-shell-core-autoconfigure/src/main/java/org/springframework/shell/core/autoconfigure/ShellRunnerAutoConfiguration.java
@@ -53,6 +53,7 @@ public class ShellRunnerAutoConfiguration {
 	}
 
 	@Bean
+	@ConditionalOnMissingClass("org.springframework.shell.jline.DefaultJLineShellConfiguration")
 	@ConditionalOnProperty(prefix = "spring.shell.interactive", name = "enabled", havingValue = "false")
 	public ShellRunner nonInteractiveShellRunner(CommandParser commandParser, CommandRegistry commandRegistry) {
 		return new NonInteractiveShellRunner(commandParser, commandRegistry);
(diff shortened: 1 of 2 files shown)


spring-shell-core-autoconfigure/src/main/java/org/springframework/shell/core/autoconfigure/ShellRunnerAutoConfiguration.java
```diff
@@ -53,6 +53,7 @@ public ShellRunner systemShellRunner(ConsoleInputProvider consoleInputProvider,
 	}
 
 	@Bean
+	@ConditionalOnMissingClass("org.springframework.shell.jline.DefaultJLineShellConfiguration")
```
fmbenhassine: That makes sense. I see we are using the same condition on the interactive shell configuration just above. So applying the same logic for the non-interactive shell seems reasonable to me.

fmbenhassine:
LGTM 👍 Rebased and merged as f701dd94732f6f4f513b039393644d1fb2f78884.

PR #1258, opened 2025-12-31
Fix NPE when script file contains empty lines at the end of file

czpilar:
Resolves #1257

PR #1320, opened 2026-02-15
Fix possible NPE in InteractiveShellRunner

czpilar:
This is a small fix of potentional **NPE** in Interactive Shell Runner.

Consider following configuration:

```java

import org.jline.reader.LineReader;
import org.springframework.boot.builder.SpringApplicationBuilder;
import org.springframework.context.annotation.Primary;
import org.springframework.shell.jline.JLineInputProvider;
import org.springframework.util.StringUtils;

@SpringBootApplication

		new SpringApplicationBuilder(SpringShellApplication.class).properties("spring.shell.debug.enabled=true")
			.run(args);
	}

	@Command(name = "hello")
	public void sayHello(@Option String name) {
		System.out.println("Hello " + name + "!");
	}

	@Bean
	@Primary
	public JLineInputProvider myInputProvider(LineReader lineReader) {
		return new JLineInputProvider(lineReader) {
			@Override
			public String readInput() {
				String s = super.readInput();
				return StringUtils.hasLength(s) ? s : null;
			}
		};
	}

}
```

Just simple enter which produces `null` in `readInput` method will end in NPE.

obrazek

### DIFF
--- a/spring-shell-core/src/main/java/org/springframework/shell/core/InteractiveShellRunner.java
+++ b/spring-shell-core/src/main/java/org/springframework/shell/core/InteractiveShellRunner.java
@@ -74,12 +74,13 @@ public abstract class InteractiveShellRunner implements ShellRunner {
 			String input;
 			try {
 				input = this.inputProvider.readInput();
-				if (input.isEmpty()) { // ignore empty lines
-					continue;
-				}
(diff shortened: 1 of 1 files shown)


czpilar:
It is similiar issue as in #1258 where it was fixed for Non Interactive Shell Runner.

PR #1321, opened 2026-02-15
Add environment post processor to determine if shell should run non-interactively when a command is present

### LINKED ISSUE #1278: Executing a shell command non-interactively is no longer possible if interactive mode is enabled in version 4.0.0
I have created a shell application with Spring Boot. In the application.yaml I have enabled interactive mode and script execution:

```
spring:
  application:
    name: AMELIOShell
  shell:
    history:
      name: "amelioshell.log"
    interactive:
      enabled: true
    script:
      enabled: true
```

With version Spring Shell 3.4 I could execute the shell non interactively by simply passing a command, e.g.:

```
java -jar .\amelioshell-5.3.0-SNAPSHOT.jar get-jobstate
```
The output of the command is displayed, and then the shell exits.

When I run the same command with a jar that was built with version 4.0 the shell starts interactively, and it stops at the prompt. If I add the parameter `--spring.shell.interactive.enabled=false` then the shell is executed non-interactively. 

This is a valid bypass for me, but the previous behaviour was much better.

Is ist possible to find out at runtime whether a command was executed interactively or non interactively?

czpilar:
Resolves #1278

### DIFF
--- a/spring-shell-core-autoconfigure/src/main/java/org/springframework/shell/core/autoconfigure/SpringShellEnvironmentPostProcessor.java
+++ b/spring-shell-core-autoconfigure/src/main/java/org/springframework/shell/core/autoconfigure/SpringShellEnvironmentPostProcessor.java
@@ -1,0 +1,31 @@
+package org.springframework.shell.core.autoconfigure;
+
+import org.springframework.boot.EnvironmentPostProcessor;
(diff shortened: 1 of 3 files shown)


fmbenhassine:
Closing this as discussed in #1278. Thank you anyway!

PR #1329, opened 2026-04-01
fix command level profile registration (#1328)

### LINKED ISSUE #1328: Spring profiles not honoured when registering commands
In the [hello world sample](https://github.com/spring-projects/spring-shell/tree/main/spring-shell-samples/spring-shell-sample-hello-world), if I add `@Profile("greetings")` to the `hello` command and start the app **_without_** activating the `greetings` profile, then the command is still registered, while it shouldn't.

This is similar to #1325 (quite the "opposite" of it) but at the command level, not the command's enclosing class level.

czpilar:
Resolves #1328


fmbenhassine:
LGTM 👍 Rebased and merged as 56cf581de30759a71ab37fc974c0b9c99ea2f7c4. I just extracted the method profile checking utility in `Utils` class (see 3a820b581bcc65e512b994991bf2b493cdfd7826).

PR #1333, opened 2026-04-02
fix boolean flag in subcommand (#1331)

### LINKED ISSUE #1331: Boolean flags not working without value in sub-commands
This is a follow-up to https://github.com/spring-projects/spring-shell/issues/1309 but it's specific to sub-commands.

# Steps taken
Consider the following application which with a command `test` with a sub-command `sub-test` which has an optional flag, i.e. a boolean argument, of name `--force`:

```java
package com.example.demo;

@SpringBootApplication
public class DemoApplication {

		SpringApplication.run(DemoApplication.class, args);
	}

	@Command(name = "test sub-test")
	public void test(@Option() boolean force) {
		System.out.println("force = " + force);
	}

}
```

# Expected behaviour

```
❯ ./mvnw -q spring-boot:run
...
$>test sub-test
force = false
$>test sub-test --force true
force = true
$>test sub-test --force
force = true
```

# Actual behaviour

⚡ `Error while parsing command: Option '--force' requires a value`
```
❯ ./mvnw -q spring-boot:run
...
$>test sub-test
force = false
$>test sub-test --force true
force = true
$>test sub-test --force
Error while parsing command: Option '--force' requires a value
```

---

_edit: I was able to confirm this on `4.0.1` and `4.0.2-SNAPSHOT` (9dbdd7180)._

czpilar:
Resolves #1331

### DIFF
--- a/spring-shell-core/src/main/java/org/springframework/shell/core/command/DefaultCommandParser.java
+++ b/spring-shell-core/src/main/java/org/springframework/shell/core/command/DefaultCommandParser.java
@@ -88,6 +88,7 @@ public class DefaultCommandParser implements CommandParser {
 		}
 		List<String> subCommands = firstNonCommandWordIndex == 1 ? Collections.emptyList()
 				: words.subList(1, firstNonCommandWordIndex);
+		String fullCommandName = firstNonCommandWordIndex == 1 ? commandName : prefix;
 
 		// add command and sub commands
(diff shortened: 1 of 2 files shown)


fmbenhassine:
LGTM 👍 Rebased and merged as 4f8bd1e426a681c393edc6890a2be61de5b8d7fb. Thank you for your contribution!

PR #1341, opened 2026-04-26
make ShellInputProvider Builder public to make it usable outside its own package

mhayen:
The current way of defining ShellInputProvider makes it impossible to use it in another package than 'org.springframework.shell.test'.


fmbenhassine:
LGTM 👍  Thank you for your contribution!

Issue #1344, opened 2026-04-27
Missing argument information in the output of the help command


bagelben:
The background to this issue is us migrating our application from Spring Boot 4 from 3.5, and along with that, moving to Spring Shell 4. Specifically Spring Boot 4.0.5 and Spring Shell 4.0.2

We have a number of commands which are similar, but I will include one of the simpler ones for reference:

In Spring Shell 3 our command was defined as follows:

```
@Command(command = "queue-depth, description = "Shows the number of messages on the specified queue", group = "Queue Depth")
    public String queueDepth(@Option(required = true, description = "Name of the Queue, or 'all' to display all queues", longNames = "queue-name") final String queueName )  {

        // Logic Goes here
    }
```

In Spring Shell 3 any Option defined can be actually be provided as an argument. This means that either of the following usages are logically equivalent

```
queue-depth queueName
queue-depth --queue-name queueName
```

We have historically relied on this before because most users would tend to use the command without providing the long option.

Now it seems like in version 4 this behaviour has deliberately removed in favour of having completely separated Options and Arguments as concepts. If there is any way of recreating the old behaviour it would be good to know. It would definitely be the most convenient for our purposes, as otherwise we will have to break one of the 2 ways of executing the command. If there is no way to achieve this we will have to migrate our Options to be Arguments, as that is our most common usage.

Now here comes the problem. Using Options is self-documenting. e.g. when we define an Option and provide a description, the help command will automatically populate it. So that the running the command `help queue-depth` would output the following useful output

```
 help queue-depth
NAME
        queue-depth - Shows the number of messages on the specified queue

SYNOPSIS
        queue-depth [--queue-name String]  --help

OPTIONS
        --queue-name String
        Name of the Queue, or 'all' to display all queues
        [Mandatory]

        --help or -h
        help for queue-depth
        [Optional]
```

Side note - it looks like the current behaviour only shows mandatory options in square brackets, and optional values are not in brackets. This feels like it is the opposite of what the behaviour should be. Although this seems to have always been the case.

Now if I was to migrate to Arguments instead of options - the following: `@Argument(index = 0, description = "Name of the Queue, or 'all' to display all queues") final String queueName` - the help would output the following:

```
 help queue-depth
NAME
        queue-depth - Shows the number of messages on the specified queue

SYNOPSIS
        queue-depth  --help

OPTIONS
        --help or -h
        help for queue-depth
        [Optional]
```

[...]

PR #1359, opened 2026-06-05
Fix missing argument information in output of help command

DF-Nico:
https://github.com/spring-projects/spring-shell/issues/1344 :

* Add Argument information to Command registration for help display
* Update help display to include @Argument
* _Update help display to have optional options (sic) between square brackets_ **(reverted)**

However, there is still a point that bugs me: @Arguments... currently the somehow sit inbetween:
- @Option: the way Arguments are handled in [MethodInvokerCommandAdapter.java](https://github.com/spring-projects/spring-shell/blob/main/spring-shell-core/src/main/java/org/springframework/shell/core/command/adapter/MethodInvokerCommandAdapter.java) is closer to Option (iterate over parameter and check annotation on the fly)
- @Argument: as they don't have name per se, it is basically an Argument too

So I think this PR should be handled maybe as a basis, and need further discussion/improvement, as currently we would have Option and Argument in help but Arguments would be missing.

IMHO Arguments should be closer to Argument and have an index stored, and maybe see @Arguments as merely an extension of Argument (or reserve, Argument is an Arguments with arity of 1)

### DIFF
--- a/spring-shell-core/src/main/java/org/springframework/shell/core/command/AbstractCommand.java
+++ b/spring-shell-core/src/main/java/org/springframework/shell/core/command/AbstractCommand.java
@@ -15,21 +15,19 @@
  */
 package org.springframework.shell.core.command;
 
+import jakarta.validation.Path;
+import org.jspecify.annotations.Nullable;
+import org.springframework.shell.core.ParameterValidationException;
+import org.springframework.shell.core.command.availability.Availability;
+import org.springframework.shell.core.command.availability.AvailabilityProvider;
+import org.springframework.shell.core.command.completion.CompletionProvider;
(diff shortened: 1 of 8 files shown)


fmbenhassine:
Thank you for the PR! I was working on the same issue and came almost to the same change set.

Let's focus on handling `@Argument` for now, `@Arguments` will be handled in a different issue/PR. Please remove any change related to options as well (namely this: "Update help display to have optional options (sic) between square brackets").

DF-Nico:
Removed the commit, should do the trick!

fmbenhassine:
Thank you for the update! LGTM now 👍 

Rebased and merged as cccc256490f97028d9d829bb74e021f0fa05a173.

PR #1360, opened 2026-06-15
Command global prefix

elysrivero99:
Today, @CommandGroup(prefix = "...") lets you namespace commands within a single class, but there is no way to apply a prefix to every command in an application at once. This is useful when embedding a Spring Shell application inside a larger system where all commands need to share a consistent namespace.

  This change introduces the spring.shell.command.prefix property as a global prefix. When set, it is prepended to the name of every user-defined @Command method. If a @CommandGroup annotation also defines a prefix on the declaring class, both are chained — global prefix first, then group prefix, then the command name. Built-in standard commands (help, clear, version, etc.) are not affected.

| Scenario | Resulting command name |
  |---|---|
  | Global `shell`, no group prefix, command `hello` | `shell hello` |
  | Global `shell`, group prefix `my-group`, command `hello` | `shell my-group hello` |
  | No global prefix, group prefix `my-group`, command `hello` | `my-group hello` |
  | No prefixes set | unchanged |


fmbenhassine:
Thanks for this, nice feature! While reviewing, I found a small functional issue when a `@Command` method has no explicit name (the common case of relying on the default method name fallback):

```java
@Command
public void unnamed() { }
// with spring.shell.command.prefix = "app"
// Result: "app " instead of the expected "app unnamed"
```

I'll follow up with a commit to address this issue.

fmbenhassine:
Rebased, squashed and merged as 7f036b151e2246fd097a848915001ec874cbc487. Refined in dc18b3d78d23894df5e5a0290ec9180dc19c08bd.

PR #1369, opened 2026-08-22
Fix implicit --help option for sub-commands

### LINKED ISSUE #1368: Incorrect handling of implicit --help option for sub-commands
I am using Spring-Shell 4.0.3. 

When defining a command via annotations the  (implicit) -help-Option does not include a default-value: 

```
@Component
public class SampleCommand {

	@Command(
		name = "foo", 
		description = "Description of 'foo'", 
		help = "Help of 'foo'")
	public void foo(
		@Argument(index = 0, description = "The desc of value") int value, 
		@Option String name) {
		System.out.println("value: " + value);
	}
	
	@Command(
		name = "bar", 
		help = "Help of 'bar'")
	public void bar(
		@Option String name) {
	}
}
```

When running `samples bar --help` I get an error message: 

```
wdc:>samples bar --help
Error while parsing command: Option '--help' requires a value
```

When running `samples bar --help true` the help will be displayed.

czpilar:
Resolves #1368

### DIFF
--- a/spring-shell-core/src/main/java/org/springframework/shell/core/command/DefaultCommandParser.java
+++ b/spring-shell-core/src/main/java/org/springframework/shell/core/command/DefaultCommandParser.java
@@ -195,12 +195,20 @@ public class DefaultCommandParser implements CommandParser {
 	}
 
 	private boolean isBooleanOption(String commandName, String currentWord) {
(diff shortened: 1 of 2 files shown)


fmbenhassine:
LGTM 👍 Thank you for your contribution!

PR #1371, opened 2026-08-22
Fix detection of the implicit help option

### LINKED ISSUE #1370: help detection ignores command's own options
`AbstractCommand` decides whether the user asked for help by looking only at the option names typed on the command line, never at what the command declares. Two defects follow from that.

```java
@Command(name = "connect", description = "Connect to a host", help = "Help of 'connect'")
public void connect(@Option(shortName = 'h', longName = "host") String host, @Option String port) {
	System.out.println("connecting to host=" + host + " port=" + port);
}
```

### 1. A declared `-h` is hijacked as help

`isHelp()` treats any option with `shortName == 'h'` as a help request. The parser resolves the value correctly and `execute()` then discards it, so the command never runs - and the exit status is `OK`, so it fails silently:

```
shell:>connect -h myhost
Help of 'connect'                     <-- expected: connecting to host=myhost

shell:>connect --host myhost
connecting to host=myhost port=null   <-- the same option, spelled long, works
```

Image

(The third line in the screenshot above is explained in the last section.)

### 2. Help is only honoured when it is the sole option

The dispatch condition is `options.size() == 1 && isHelp(...)`, so anything else on the line suppresses it:

```
shell:>connect --help=true
Help of 'connect'

shell:>connect --help=true --port 22
connecting to host=null port=22       <-- expected: Help of 'connect'
```

Image

### These cannot be fixed separately

`options.size() == 1` is what currently masks defect 1 (the third line of the first screenshot):

```
shell:>connect -h myhost --port 22
connecting to host=myhost port=22     <-- correct today, only because size() == 2
```

Relax the dispatch on its own and that line prints help instead, widening defect 1 from one option to any number. A change fixing only defect 2 is therefore a regression.

Reproduced on `main`.

czpilar:
Resolves #1370

`isHelp()` now ignores options the command declares itself, so a declared `-h` is no longer hijacked as help. The dispatch condition is relaxed from `options.size() == 1` to `anyMatch`, so help is honoured alongside other options - the two defects cannot be fixed separately, see the issue for why.

### Case sensitivity

`isHelp()` also matched the long form case-insensitively, unlike every other option comparison in the codebase. This PR changes it to `"help".equals(...)`, so `--HELP` is no longer treated as help. Repro and details in #1370.

The alternative - making the long form consistently case-insensitive - would imply `-H` is help, and `-H` is an established distinct flag elsewhere (`curl -H`, `grep -H`, `ls -H`), which is exactly the hijack this PR fixes. Nothing documents the uppercase spelling: the built-in help prints `--help or -h` for every command, and the reference docs mention only
`--help` / `-h`.

### Note on the dispatch condition

`commands/help.adoc` already states that help short-circuits execution "regardless what other command-line options are typed", so relaxing `options.size() == 1` to `anyMatch` brings the code in line with documented behaviour rather than changing it.

### DIFF
--- a/spring-shell-core/src/main/java/org/springframework/shell/core/command/AbstractCommand.java
+++ b/spring-shell-core/src/main/java/org/springframework/shell/core/command/AbstractCommand.java
@@ -23,6 +23,7 @@ import org.springframework.shell.core.command.availability.AvailabilityProvider;
 import org.springframework.shell.core.command.completion.CompletionProvider;
 import org.springframework.shell.core.command.completion.DefaultCompletionProvider;
 import org.springframework.shell.core.command.exit.ExitStatusExceptionMapper;
+import org.springframework.util.StringUtils;
 
 import java.io.PrintWriter;
 import java.util.ArrayList;
@@ -35,6 +36,7 @@ import java.util.Objects;
  * @author Janne Valkealahti
  * @author Piotr Olaszewski
  * @author Mahmoud Ben Hassine
+ * @author David Pilar
  */
 public abstract class AbstractCommand implements Command {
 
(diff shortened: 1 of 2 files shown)


fmbenhassine:
Rebased and merged as 1918c0df7fde97c60839be783e271b1b4384d83b.

PR #1372, opened 2026-08-22
Report variadic arguments in the help message

### LINKED ISSUE #1367: Missing arguments information in help message of commands
This is an addition to the already closed issue #1344 : 

I use the current version of spring-shell 4.0.3. It seems that the issue 1344 has been resolved but I miss the documentation for @Arguments. Look at the following example: 

```
@Component
public class SampleCommand {

	@Command(
		name = "foo", 
		description = "Description of 'foo'", 
		help = "Help of 'foo'")
	public void foo(
		@Argument(index = 0, description = "The desc of value") int value, 
		@Option String name) {
		System.out.println("value: " + value);
	}
	
	@Command(
		name = "bar", 
		help = "Help of 'bar'")
	public void bar(
		@Option String name) {
	}	
}
```

When I now call `help samples bar` the output of the help message does not mention anything about its arguments: 

```
NAME
        samples bar - Description of 'var'

SYNOPSIS
        samples bar --name String --help

OPTIONS
        --name String

        [Optional, default = ]

        --help or -h
        help for samples bar
        [Optional]
```

It seems that the @Arguments-Annotation does not have an "description" and that the information about @Arguments is completely ignored in the `help`-output.

czpilar:
Resolves #1367

### DIFF
--- a/spring-shell-core/src/main/java/org/springframework/shell/core/command/CommandArgument.java
+++ b/spring-shell-core/src/main/java/org/springframework/shell/core/command/CommandArgument.java
@@ -22,10 +22,11 @@ import org.jspecify.annotations.Nullable;
  * argument.
  *
  * @author Mahmoud Ben Hassine
+ * @author David Pilar
  * @since 4.0.0
  */
 public record CommandArgument(int index, @Nullable String description, @Nullable String defaultValue,
-		@Nullable String value, Class<?> type) {
+		@Nullable String value, Class<?> type, boolean variadic) {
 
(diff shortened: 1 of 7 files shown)


tschoellhorn:
This is great. Thank you for looking into it!

fmbenhassine:
LGTM 👍 Rebased and merged as e261efee2c9e633fe7d229299f0e5a3a9278293e. I also updated the docs in 70c116346c06f79cee37e61a9e100b820480a4ba.

Thank you for all your contributions!

PR #1384, opened 2026-09-21
Fix inverted SYNOPSIS brackets in help output

### LINKED ISSUE #949: required Boolean with non-default value
**spring shell version: 3.1.6**

In one of my commands, I want to make sure that the end-user specifically indicates whether they either want A or B by using a Boolean. I don't want to make the choice for the end user.

When I add the following @Option on my command:

```java
@Option(required = true, description = "Description", longNames = "Some long name")  Boolean booleanOption
```

Spring Shell, automatically transforms the booleanOption to defaultValue `false`. The help menu of the command indicates that the booleanOption is optional. 

When transforming into

```java
@Option(required = true, description = "Description", longNames = "Some long name")  boolean booleanOption
```

Still, the help menu indicates the booleanOption as optional. I assume for 2 reasons that this should not be the case: (1) it is a boolean and not Boolean. (2) required is explicitly set to true. The defaultValue is again false.

Setting the defaultValue to 'true' explicitly seems to be working.

```java
@Option(required = true, description = "Description", longNames = "Some long name", defaultValue = "true")boolean booleanOption
```

It will show that the boolean is optional, but the default value will be true.

But again, I didn't want to choose for the user. I want the end user to articulate their choice when doing a command specifically. 

If I write:

```java
@Option(required = true, description = "Description", longNames = "Some long name", defaultValue = "")Boolean booleanOption
```

Or 

```java
@Option(required = true, description = "Description", longNames = "Some long name", defaultValue = "null")Boolean booleanOption
```
The defaultValue still translates to false (while as a Boolean I would expect it to be able to be null).

Is this the intended behavior?

ezeprimo:
Fixes #949

Required options were rendered as `[--opt Type]` and optional ones as bare text, opposite to Unix convention and to the arguments logic in the same file. This made required options (e.g. a required Boolean) look optional in the help.

Summary:
- Render required options bare (`--opt Type`) and optional ones bracketed (`[--opt Type]`) in SYNOPSIS
- Updated the HelpTests expectations that pinned the old output

| File | Change |
|------|--------|
| `spring-shell-core/.../command/Help.java` | Un-inverted SYNOPSIS brackets (2 lines) |
| `spring-shell-core/.../command/HelpTests.java` | Updated 5 synopsis expectations |

Test plan:
- `mvn -pl spring-shell-core -am -Dtest=HelpTests test`: 6/6 green (JDK 17)
- `mvn -pl spring-shell-core -am test`: 217/217 green


fmbenhassine:
Thank you for the PR. I don't see how it is related to #949 .

But it actually fixes a valid inconsistency, I created an issue for it: #1389.

Please rebase this PR on the latest `main` as the help rendering moved to  `CommandHelpRenderer` (which still has the inverted logic today) and it should be good to merge.

PR #1385, opened 2026-09-21
Render the same help message for --help and help <command>

### LINKED ISSUE #1373: Inconsistent help message between "mycommand --help" and "help mycommand"
I use Spring Boot 4.0.3. 

When defining a Command with the @Command-Annotation it is possible to set a "help"-text. Currently this text is displayed when using the "--help" option. But: This text is displayed without parameters, options and synopsis. Consider this case: 

```
@Component
public class SampleCommand {
	@Command(
		name = "bar", 
		help = "Help of 'bar'")
	public void bar(
		@Option String name) {
	}
}
```

When running `samples bar --help` the output is:

```
Help of 'bar'
```

But when running `help samples bar` this output is produced: 

```
NAME
        samples bar - Description of 'bar'

SYNOPSIS
        samples bar --name String --help

OPTIONS
        --name String

        [Optional, default = ]

        --help or -h
        help for samples bar
        [Optional]
```

In the later variant the text of the @Command.help is completely ignored. I suggest the following: 

1.) The output of both "help" variants should be same - the base for that could be the `help samples bar` version.
2.) The @Command.help text should be integrated in that output. 

I saw other issues which implicitely mentioned the same idea. This would streamline the help output.

czpilar:
Resolves #1373


fmbenhassine:
That's great 👍 Thank you for your contribution!

Just a couple minor changes:
- I would rename `render(Command)` to `renderHelp(Command)`. In fact, when imported statically, `render(myCommand)` could be confused with rendering the command's result or something else. Being explicit is better I think
- Since you introduced a new class for help rendering, I will add a separate class for its unit tests.

I will take care of that on merge.

fmbenhassine:
Merged as aa892f8ee1e5ff2bb3c1422ed6d0c5acbbe0455d, refined as discussed in 8184e97a9008fdf770416b1934af524159895633.

Thanks again for all your contributions to the project. Really appreciated 🙏

PR #1386, opened 2026-09-21
Allow script command to load from classpath and file resources

### LINKED ISSUE #1014: Extend "script" command to accept classpath references
The script command does accept files only at the moment.

If the command would use something like this:

```
public void script(String ref) throws Exception {
        Resource res = getApplicationContext().getResource(ref);
        Reader reader = new StringReader(new String(res.getInputStream().readAllBytes(), StandardCharsets.UTF_8));
        try (FileInputProvider inputProvider = new FileInputProvider(reader, parser)) {
            getShell().run(inputProvider);
        }
}
```
the script could be something like: "classpath:scripts/runthis" or "file:...." - all things possible which the app context can resolve.
To be backwards compatible you could prefix "ref" always with "file:.." if nothing else is specified on the argument.

ezeprimo:
Fixes #1014

`script --file` only accepted absolute filesystem paths. It now resolves the value with `DefaultResourceLoader`, so `classpath:...` and `file:...` references work. An existing filesystem path keeps the previous behavior (backward compatible, no new options).

Summary:
- `Script` resolves `--file/-f` as a Spring resource, file-first fallback
- `FileInputProvider` gained a `Reader` overload reusing the line-continuation/comment logic

| File | Change |
|------|--------|
| `spring-shell-core/.../command/Script.java` | Resource resolution with file-first fallback |
| `spring-shell-core/.../FileInputProvider.java` | New `Reader` constructor |
| `spring-shell-core/.../command/ScriptTests.java` | New: classpath, absolute path, file-URL, missing-file |
| `spring-shell-core/.../FileInputProviderTests.java` | New Reader test |

Test plan:
- `mvn -pl spring-shell-core -am -Dtest=ScriptTests,FileInputProviderTests test`: 6/6 green (JDK 17)
- `mvn -pl spring-shell-core -am test`: 222/222 green


spring-shell-core/src/test/java/org/springframework/shell/core/command/ScriptTests.java
```diff
@@ -0,0 +1,121 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.CleanupMode;
+import org.junit.jupiter.api.io.TempDir;
+
+import org.springframework.shell.core.InputReader;
+
+/**
+ * Tests for {@link Script}.
+ *
+ * @author David Pilar
```
czpilar: You should use your own name as an author, not mine... ;-)

czpilar (changes requested):

ezeprimo:
Author adjusted.

fmbenhassine:
Thank you for the contribution! LGTM, I will merge it. However, I will add a follow-up commit to restrict resource resolution to `classpath:` and `file:` prefixes only (as the original OP in #1014 requested). The reasoning is that even though `script` only ever executes commands already registered in the app, the resource loading step itself (as it is currently implemented in this PR) can be pointed at other URL schemes (like `http:` or `https:`), which would make the application issue a network call to a location supplied via the `--file` option. Restricting to `classpath:`/`file:` closes that (security gap) off while resolving #1014 as requested.

fmbenhassine:
Rebased, squashed and merged as 07a0e667ad1bb5909465efe370a4a0c3fb7c1539. Refined in 7fb993ba542373bf76a13c0b63943364116fde7e.

ezeprimo:
Thanks to you! I see the refinements were necessary — excellent review. My PR didn't consider those cases, and they improved security.

